PR #11, opened 2023-03-06
Rename DeferredLayoutEffectsProvider to LayoutGroup

### LINKED ISSUE #9: Deferred layout effects are poorly named
I still find "deferred" to be non-obvious naming. While it's true that using `useDeferredLayoutEffect` registers an effect that runs later than `useLayoutEffect`, it doesn't say anything about when "later" is.

I've been thinking for a while that `<LayoutGroup />` and `useLayoutGroupEffect` might make more sense. The `<LayoutGroup />` components creates a boundary in which layout effects can be grouped together and processed after all the children of the group have run their layout effects by way of `useLayoutGroupEffect`.

SMores:
Closes #9.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -8,9 +8,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v3
         with:
           node-version: 18.x
 
--- a/.yarn/versions/87656097.yml
+++ b/.yarn/versions/87656097.yml
@@ -1,0 +1,2 @@
+declined:
+  - "@nytimes/react-prosemirror"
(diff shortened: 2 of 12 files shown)


.github/workflows/check.yml
```diff
@@ -8,9 +8,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
SMores, line 13: This stinks a little, but in order for `yarn version check` to work, we need at least the git history from `main` to `HEAD`. I'd much rather _only_ fetch exactly those commits (the default behavior is to fetch only a single commit, the branch tip), but I have not yet figured out a good way to do that :P. If anyone has any ideas, lemme know!

.yarn/versions/87656097.yml
```diff
@@ -0,0 +1,2 @@
+declined:
+  - "@nytimes/react-prosemirror"
```
SMores, line 2: I think this is fine; this is a breaking change, but we haven't actually published this anywhere yet!

.github/workflows/check.yml
tilgovi: 🤷🏻 I wouldn't stress about it for now. I don't expect the change history here will get huge or have huge binaries or anything.

tilgovi (approved):

annabialas (approved):

PR #12, opened 2023-03-06
RFC: Implement a synchronous React-NodeView system

SMores:
This draft PR is an experiment in seeing how far we can get with a synchronous React-NodeView system. The basic idea is to drop the portal system, and instead rely on a jerry-rigged react-dom renderer that's been configured with `isPrimaryRenderer: false`, allowing it to be synchronously rendered (with `flushSync(() => { root.render(element) })`) from within a standard react-dom render cycle.

This is not particularly likely to get merged into `main`. It has several downsides:

* NodeView components are no longer rendered within the primary React tree; each one is back to its own React tree again. This is actively a step backwards to what we had before the portal system.
* ~Context doesn't flow automatically from one renderer to another, so there needs to be a manual binding point (something I'm still figuring out). Again, this is a step backwards from the portal system~ Latest WIP commit attempts to resolve this
* It relies on this janky renderer, though if we decide actually we love this and we want to move forward with it, we can probably make this less janky

### Trying to add context bridging

Ok so this second commit is doing some _truly bonkers_ stuff (probably not all of it necessary, this was very stream-of-consciousness).

One thing of note here is that it is maybe impossible to actually bridge the EditorViewContext across the renderer gap, because you need the nodeViews to get the EditorView, and you need the EditorView to provide the EditorViewContext, so there's a circular dependency. Buuuut this maybe doesn't matter so much, because the EditorViewContext mostly just provides the EditorView, and that gets directly passed in to the nodeviewconstructor, so we can just render the context manually!


tilgovi:
> NodeView components are no longer rendered within the primary React tree; each one is back to its own React tree again. This is actively a step backwards to what we had before the portal system.

Not quite, right? They're not part of the same tree, just using multiple roots within the secondary renderer. How do they present in the dev tools?

SMores:
I'm not exactly sure what you're saying, but this is what it looks like in dev tools:

image

SMores:
Is there a difference between "multiple roots" and "multiple trees"? Is each root not the beginning of a separate tree?

PR #15, opened 2023-03-09
0.1.1

SMores:
Turns out, `yarn version apply` is a little broken! There are some kinks to be worked out for sure. Some things that I've learned:

The `--all` flag is meant to only apply to monorepos; in theory, in a single-project repo, it shouldn't be necessary. In practice, if you don't include `--all`, `yarn version apply` doesn't actually delete any version files, it just clears out the contents of any (except for `declined` bumps, which it doesn't seem to know how to handle). To work around this, we just always should use `--all` (as noted in the new Version Bump docs), which works fine on single-project repos.

Additionally, `yarn version check` kind of explodes if you run it after having run `yarn version apply --all`, because `yarn version check` kind of explodes if you modify more than one version file in a branch, and `yarn version apply --all` deletes all of the version files! To work around _this_, I introduced a new, dumb bash script that only runs the command you pass it if `package.json`'s version field hasn't been updated on this branch. This way we can simply run `./scripts/if-no-version-change.sh yarn version check` in our pre-push hook to avoid running `yarn version check` if `yarn version apply --all` has already been run!

Finally, of course, this PR actually bumps the version to 0.1.1!

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -29,4 +29,6 @@ jobs:
         run: yarn check:types
 
       - name: Version
-        run: ./scripts/if-no-version-change.sh yarn version check
+        run:
+          git fetch origin main --depth 1 && ./scripts/if-no-version-change.sh
+          yarn version check
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -29,4 +29,4 @@ jobs:
         run: yarn check:types
 
       - name: Version
-        run: yarn version check
+        run: ./scripts/if-no-version-change.sh yarn version check
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn version check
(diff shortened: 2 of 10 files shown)


tilgovi:
Since we're not doing any continuous deployment and this is a single package, should we just be bumping the version number manually when it's time to release, instead, without worrying about staging version changes?

SMores:
> should we just be bumping the version number manually when it's time to release, instead, without worrying about staging version changes?

I think the thing that's nice about this is that we don't have to look back when it's time to release and figure out what kind of version bump based on the who-knows-how-many changes that we've done since we last released, because we're documenting each change instead.

tilgovi:
Wanna could try [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).

SMores:
We could! But that doesn't seem like a particularly better way to do it to me, just a different one

tilgovi:
Fair. Just wanted to give you options yarn was proving difficult.

SMores:
Thanks @tilgovi :) Yeah this _feels_ like it should be stable now, but I agree that if we find ourselves needing to mess with this more, we should consider other options

tilgovi (approved):

PR #17, opened 2023-03-17
Add a hook for registering EditorView-level keyboard events

SMores:
## Premise

There are certain events that can _only_ be properly received and handled at the ProseMirror View level. For example, if you want to change the behavior of a certain keyboard event that ProseMirror would normally care about, like pressing the Enter key, you can _only_ prevent ProseMirror from executing its default behavior by specifying a `handleKeyDown` prop, either in the direct editor props or in a plugin. This is because browsers don't emit keyboard events for elements that are children of a contenteditable, even if that happens to be where the selection is when the keyboard event is generated. The event target is always the contenteditable itself, not the child element.

You can't even straightforwardly use React event handlers to listen for mouse events on ProseMirror NodeViews. React actually puts all event handlers on the root, rather than adding them to individual elements, which means that ProseMirror will "hear" the user's click event first, and have a chance to handle it before it even gets to React. Your event handler still gets called for the event, eventually, but it can't cancel the default ProseMirror behavior, because ProseMirror has already finished responding!

Sometimes this is fine; if you want some sort of generalized behavior for all nodes or all nodes of a given type, you can simply modify the default key bindings or chain additional commands.

However, if you have written your NodeViews as React components, and wish to co-locate your state management logic for a given NodeView within that component (and therefore alongside the view rendering logic), you're sort of stuck. Your only real option is to separate the logic, storing the state management in a plugin, and the view rendering in a component.

## Proposal

The introduction of a new hook (name TBD, currently called `useProseMirrorEvent` which... meh) that allows NodeView components to register event handlers that are then ("behind-the-scenes") executed by a ProseMirror Plugin.

### Challenges

Ok so first thing is that naming is just going to be a challenge. This hook is essentially the inverse of `useEditorEvent`; instead of making the EditorView available to event handlers that are associated with React components, this hook will allow React component state to be accessed in a ProseMirror event handler.

Also, I'm not convinced that requiring components to check whether the selection is in their node is like... reasonable, or too annoying? Other alternatives might include passing the node or a ref to the dom into the hook, in addition to the handler, but that feels kinda clunky as well.


tilgovi:
I see the need for a way to tap into ProseMirror's event handling, because React's event handling doesn't kick in until the events bubble all the way to the React root. It's also super convenient to be able to refer to a node view's ProseMirror node for things like the auto-complete. In the auto-complete example, though, do we actually need to register the handler at the ProseMirror view level, or would it be simpler to attach a handler directly to the view's DOM node? It seems like if a handler on the DOM node were to call `stopPropagation()` then ProseMirror would never handle it. This handler would have to be added with `addEventHandler`, rather than React.

SMores:
Ahhh, I had forgotten that _that_ (React's event handlers being attached to the root) was the reason this is a pain! Thank you for pointing that out, @tilgovi!

> In the auto-complete example, though, do we actually need to register the handler at the ProseMirror view level, or would it be simpler to attach a handler directly to the view's DOM node? 

That's a great point/question. I suspect simply using a ref and manually attaching an event handler would be sufficient; doing so would certainly be simpler if it works. Let me sketch out a version of that and see what it would look like.

SMores:
Lol, right. That's _not_ the entire reason that this is a pain. In this case, you would need to at least put the event listener on the EditorView.dom, because paragraph tags don't emit keydown events, even within contenteditables; only the actual contenteditable does that!

I just double checked, and it looks like even putting the event listener on `view.dom` in, say, a `useEditorEffect`, doesn't let you do this with just standard event listeners, because ProseMirror's EditorView/plugin event listeners are _also_ on `view.dom`, so by the time your handler's code runs `event.stopPropagation()`, ProseMirror's already done!

SMores:
I elaborated a little bit on why this is specifically challenging in the PR description!

tilgovi:
> because paragraph tags don't emit keydown events, even within contenteditables

Right, right. I suppose you could make the `dom` not editable and make the `contentDOM` editable again, but that workaround feels hacky in that the fact these aren't the same `Node` is just an awkward limitation of this library.

Alright then. Can we use `handleDOMEvents` to make this not keyboard specific?

SMores:
Yes, for sure!

SMores:
Hm, yeah fair enough. I can give that a shot!

SMores:
@tilgovi would love your thoughts on this approach (now with `handleDOMEvents`! This does the minimum amount of plugin reconfiguration (I think!). Could also use some help with naming :)

src/contexts/EditorDOMEventRegistryContext.tsx
```diff
@@ -0,0 +1,107 @@
...
+  );
+
+  const reactEventPlugin = useMemo(
+    () => createReactEventPlugin(registry),
+    // We force an update whenever the set of event types that
+    // we care about changes. This plugin only needs to be
+    // recreated when that set (the set of registry keys)
+    // changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [registry.keys()]
```
tilgovi, line 66: This returns a new iterable every time. We could spread it, because the order should be stable, but maybe we can just do something more obvious. For example, instead of `forceUpdate`, we could the registry in state and set it to a new object when we add/remove keys (but mutate its values in place when we don't).

demo/main.tsx
```diff
@@ -22,8 +27,116 @@ const editorState = EditorState.create({
...
+      return false;
+    },
+    [stringAfterAt, highlightIndex, namesToDisplay]
+  );
+
+  return (
+    <>
+      <p ref={ref}>{children}</p>
+      {namesToDisplay && (
+        <ul contentEditable={false}>
```
tilgovi, line 123: Minor nit, because I know this is a demo, but does this generate a warning? My memory is that it doesn't matter what you set `contentEditable` to, React warns if you set it at all. If so, maybe add the suppression prop, too.

package.json
```diff
@@ -52,6 +52,7 @@
     "@typescript-eslint/parser": "^5.51.0",
     "@vitejs/plugin-react": "^3.1.0",
     "@yarnpkg/sdks": "^3.0.0-rc.38",
+    "chokidar": "^3.5.3",
```
tilgovi, line 55: Was this an optional dependency of some tool's watch mode or something? I don't see it called explicitly.

src/plugins/createReactEventPlugin.ts
```diff
@@ -0,0 +1,31 @@
...
+> = (view: EditorView, event: DOMEventMap[EventType]) => boolean | void;
+
+export function createReactEventPlugin(
+  eventHandlerRegistry: Map<keyof DOMEventMap, Set<EventHandler>>
+) {
+  const domEventHandlers: Record<keyof DOMEventMap, EventHandler> = {};
+  for (const [event, handlers] of eventHandlerRegistry.entries()) {
+    function handleEvent(view: EditorView, event: Event) {
+      for (const handler of handlers) {
+        if (handler(view, event)) return true;
```
tilgovi, line 19: Should we align with ProseMirror's own handling of this and return `true` if the handler returns true or if the handler prevented the default action?

https://github.com/ProseMirror/prosemirror-view/blob/master/src/input.ts#L78

tilgovi: That'd be simulating what it's like to register a separate plugin for each invocation.

tilgovi (commented):
For naming, what about `useEditorEventListener`? I came to that by looking at what `ahooks` has for their DOM event listener hook.

demo/main.tsx
tilgovi: I see your comment about not leaving the demo code here and resolve this.

src/plugins/createReactEventPlugin.ts
tilgovi, line 19: Should we be using `useEditorEvent` under the hood to make the event handler reference stable while capturing other values in closures that are consistent with the current view state?

src/contexts/EditorDOMEventRegistryContext.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const editorDOMEventRegistry = useMemo(
+    () => ({ register, unregister }),
+    [register, unregister]
+  );
+
+  return { editorDOMEventRegistry, reactEventPlugin };
+}
+
+type Register<EventType extends keyof DOMEventMap = any> = (
```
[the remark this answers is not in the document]
tilgovi: Do you need a default for the template parameter?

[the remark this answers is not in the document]
tilgovi: I notice you don't have it on `useEditorDOMEvent`.

src/contexts/EditorDOMEventRegistryContext.tsx
SMores: .. whoops. Yes, lemme try that!

src/contexts/EditorDOMEventRegistryContext.tsx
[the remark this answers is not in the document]
SMores: Basically, yes, I _think_. Or at least, I do at _some_ point, because at some point I have to pass this type to `createContext`, and I don't think I can have the result of `createContext` be a parameterized type?

I don't need it on `useEditorDOMEvent` because it can be inferred from usage (that's the whole goal, really, is to support type inference of the `event` parameter based on the `eventType`)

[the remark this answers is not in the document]
tilgovi: Is there any value in using keyof DOMEventMap here, or should we just use `string`?

SMores:
> For naming, what about `useEditorEventListener`?

Maybe. What I'm worried about is distinguishing this hook from `useEditorEvent`. They're very similar in their API _and_ effect, and I'm worried that it won't be clear when to use one vs the other

src/contexts/EditorDOMEventRegistryContext.tsx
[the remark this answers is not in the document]
SMores: Yes; the value is that we can correlate the types, so that the event parameter has the correct type! For example, in the demo in this branch, because the eventtype is "keydown", the `event` is typed as `KeyboardEvent`. This is the same type correlation that ProseMirror does for `handleDOMEvents`

package.json
SMores: Yeah, `@swc/cli`. I don't actually know why I added this, though, because we don't have any scripts that use swc's watch mode; everything seems fine if I remove it.

src/contexts/EditorDOMEventRegistryContext.tsx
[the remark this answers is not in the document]
tilgovi: Ah, okay. Can we use `= keyof DOMEventMap` rather than `= any` then?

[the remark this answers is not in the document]
SMores: No :P there's some covariance/contravariance thing that I'm having trouble wrapping my head around, but if I default to keyof DOMEventMap, then I get an error at the actual callsite in the `useEditorDOMEvent` hook that EventType could be instantiated with a different subtype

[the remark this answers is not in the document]
tilgovi: I think maybe what's happening here is that you have a generic type whose instantiations are non-generic functions rather than generic function types. I don't know if that actually makes sense, but am I wrong or does this seem to work?

```diff
diff --git a/src/contexts/EditorDOMEventRegistryContext.tsx b/src/contexts/EditorDOMEventRegistryContext.tsx
index f69d975296c4..5ace8df2e0f8 100644
--- a/src/contexts/EditorDOMEventRegistryContext.tsx
+++ b/src/contexts/EditorDOMEventRegistryContext.tsx
@@ -74,18 +74,16 @@ export function useReactEventPlugin() {
   return { editorDOMEventRegistry, reactEventPlugin };
 }
 
-type Register<EventType extends keyof DOMEventMap = any> = (
-  eventType: EventType,
-  handler: EventHandler<EventType>
-) => void;
-
-type Unregister<EventType extends keyof DOMEventMap = any> =
-  Register<EventType>;
-
-type EditorDOMEventRegistryContextValue = {
-  register: Register;
-  unregister: Unregister;
-};
+interface EditorDOMEventRegistryContextValue {
+  register<EventType extends keyof DOMEventMap>(
+    eventType: EventType,
+    handler: EventHandler<EventType>
+  ): void;
+  unregister<EventType extends keyof DOMEventMap>(
+    eventType: EventType,
+    handler: EventHandler<EventType>
+  ): void;
+}
 
 const EditorDOMEventRegistryContext = createContext(
   null as unknown as EditorDOMEventRegistryContextValue
```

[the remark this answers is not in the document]
SMores: Ohhhhhhhh... yes I bet this will work. Gonna try it. Ugh I should have known I literally just had to do this in a different project this morning!!!

annabialas:
@SMores i just got pinged to leave a review - would u still like one? i can look tomorrow :-)

README.md
```diff
@@ -253,18 +255,60 @@ export function ProseMirrorEditor() {
...
+components as event handlers. If you need to listen to events that originate
+_within the `contenteditable` node_, however, those event listeners need to be
+registered with the `EditorView`'s `handleDOMEvents` prop.
+
+You can use the `useEditorEventListener` hook to accomplish this. It takes an
+`eventType` and an event listener. The event listener follows the usual
+semantics for ProseMirror's `handleDOMEvents` prop:
+
+- Returning `true` or calling `event.preventDefault` will prevent other
+  listeners from running
```
tilgovi, line 270: Minor nit: missing period

README.md
```diff
@@ -253,18 +255,60 @@ export function ProseMirrorEditor() {
...
+- Returning `true` will not automatically call `event.preventDefault`; if you
+  want to prevent the default contenteditable behavior, you must call
+  `event.preventDefault`.
+
+You can use this hook to implement custom behavior in your NodeViews:
```
tilgovi, line 275: super minor nit: I think I've been saying "`NodeView`" when talking about the class and "node views" when talking about the concept.

README.md
```diff
@@ -438,17 +490,17 @@ Creates, mounts, and manages a ProseMirror `EditorView`.
 
 All state and props updates are executed in a layout effect. To ensure that the
 EditorState and EditorView are never out of sync, it's important that the
-EditorView produced by this hook is only accessed through the `useEditorEvent`
-and `useEditorEffect` hooks.
+EditorView produced by this hook is only accessed through the
```
tilgovi, line 493: We should add `useEditorEventListener` to this list, too, yeah?

tilgovi: Or we could say "through the hooks that this library exposes" or something.

README.md
tilgovi: Leave it alone, actually. I think I'll take a pass through the whole docs with this in mind after this. I see we've got this in a bunch of places.

src/components/ProseMirrorInner.tsx
```diff
@@ -38,15 +39,34 @@ export function ProseMirrorInner({
   mount,
   ...editorProps
 }: ProseMirrorProps) {
+  const { reactEventPlugin, registerEventListener, unregisterEventListener } =
+    useReactEventPlugin();
```
tilgovi, line 43: If this is feeling verbose, you could grab the `reactEventPlugin` and capture the rest (e.g. `...eventContextValues`) to spread in below.

src/hooks/useEditorView.ts
```diff
@@ -87,7 +87,9 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
   }, [mount, props, view]);
 
   useLayoutEffect(() => {
+    console.log(nonStateProps.plugins);
     view?.setProps(nonStateProps);
+    console.log(view?.props.plugins);
```
tilgovi, line 92: 🗑️

src/hooks/useReactEventPlugin.tsx
```diff
@@ -0,0 +1,51 @@
+import type { DOMEventMap } from "prosemirror-view";
+import { useCallback, useMemo, useState } from "react";
+
+import { createReactEventPlugin } from "../plugins/createReactEventPlugin.js";
+import type { EventHandler } from "../plugins/createReactEventPlugin.js";
+
+/**
+ * Produces a plugin that can be used with ProseMirror to handle DOM
+ * events at the EditorView.dom element.
```
tilgovi, line 9: better as "at the editor mount point" or nah?

src/hooks/useReactEventPlugin.tsx
```diff
@@ -0,0 +1,51 @@
...
+    new Map<keyof DOMEventMap, Set<EventHandler>>()
+  );
+
+  const registerEventListener = useCallback(
+    (eventType: keyof DOMEventMap, handler: EventHandler) => {
+      const handlers = registry.get(eventType) ?? new Set<EventHandler>();
+      handlers.add(handler);
+      if (!registry.has(eventType)) {
+        // We only need to force an update when the set of event types
+        // that we care about changes
```
tilgovi, line 29: This comment still makes sense, I think, but "force an update" makes less sense without `forceUpdate`. May or may not need the comment at all.

src/index.ts
```diff
@@ -1,10 +1,14 @@
...
+export { ProseMirror } from "./components/ProseMirror.js";
+export { EditorProvider } from "./contexts/EditorContext.js";
+export { LayoutGroup, useLayoutGroupEffect } from "./contexts/LayoutGroup.js";
+export { useEditorEffect } from "./hooks/useEditorEffect.js";
+export { useEditorEventCallback } from "./hooks/useEditorEventCallback.js";
+export { useEditorEventListener } from "./hooks/useEditorEventListener.js";
+export { useEditorState } from "./hooks/useEditorState.js";
+export { useEditorView } from "./hooks/useEditorView.js";
+export { useNodeViews } from "./hooks/useNodeViews.js";
+export { useReactEventPlugin } from "./hooks/useReactEventPlugin.js";
```
tilgovi, line 12: Yeah, I like this better. Thanks.

src/plugins/createReactEventPlugin.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const plugin = new Plugin({
+    key: new PluginKey("reactEvents"),
+    props: {
+      handleDOMEvents: domEventHandlers,
+    },
+  });
+
+  return plugin;
+}
```
tilgovi, line 34: Maybe we should inline this file into `useReactEventPlugin`.

tilgovi (commented):
Looking pretty great. I'm still contemplating `useReactEventPlugin`, and in particular whether "React event" implies the use of `SyntheticEvent`, or at least batched updates.

I'm thinking about ways to get `SyntheticEvent` but everything I can come up with is tricky and/or foolish.

tilgovi:
Maybe we can come up with a naming tweak, even though I know `useReactEventPlugin` doesn't need to be public API. And maybe we should consider running the handlers in a batched update context.

src/hooks/useReactEventPlugin.tsx
SMores: I genuinely don't know haha. What do you think? I think both are going to be equally confusing, initially

SMores:
@tilgovi I took a stab at batching the event listener calls and an alternative name for the plugin hook. I think I'm not seeing the value that SyntheticEvents would add here, even if was straightforward to do so

tilgovi:
The value would be small. Browser's event objects have converged pretty well to spec these days.

src/hooks/useEditorEventListener.ts
```diff
@@ -1,9 +1,10 @@
+import { Plugin } from "prosemirror-state";
```
tilgovi, line 1: type-only import?

src/hooks/useEditorEventListener.ts
```diff
@@ -24,12 +25,14 @@ export function useEditorEventListener<EventType extends keyof DOMEventMap>(
...
-      return ref.current(view, event);
-    },
-    []
-  );
+  const eventHandler = useCallback(function (
+    this: Plugin,
+    view: EditorView,
+    event: DOMEventMap[EventType]
+  ) {
+    return ref.current.call(this, view, event);
```
tilgovi, line 33: Nice change. Users probably won't try to use `this`, but good to be consistent.

tilgovi (approved):
I'm feeling good about this!

PR #25, opened 2023-04-27
Nest node views according to node hierarchy

SMores:
This change moves the responsibility for rendering node view components to their parents. To support this, the portal registry has been updated to be a map from "registration key" (the same random strings that we use as React keys) to an array of portals.

Each node view now registers a portal under its parent's key. Parents then retrieve the list of child portals under their key, allowing portals to be rendered with the appropriate hierarchy.

The result is that the React component hierarchy for node view components actually matches the node and dom hierarchy! Previously, all node view components would be direct children of the ProseMirror component, and all siblings of each other. This change means that node view components can now consume context from their parent node view components, which was impossible before!

image


tilgovi:
There's a potential optimization opportunity here that we discussed outside GitHub. We might be able to make portal mount/unmount cause state updates lower in the tree if we can figure out how to move the registry map values into deeper component state. It's not yet clear how to do that, or if it's possible to do it. I'm stating this opportunity here for awareness, in the hope that we might have a clever idea. What we've got here isn't worse than what we have, though, so this idea is non-blocking.

src/contexts/PortalTreeContext.ts
```diff
@@ -0,0 +1,22 @@
+import { ReactPortal, createContext } from "react";
+
+/**
+ * The "Portal Tree" is a registry of React Portals, indexed by
+ * the "key" of their parent node views.
+ *
+ * Each node view registers a portal with the tree under its
+ * parent's key. Parents then retrieve the list of child portals
+ * under their key, allowing portals to be rendered with the
+ * appropriate hierarchy.
```
tilgovi, line 10: Would this read better with the focus inverted?

> Each React Node View creates a unique key under which its React Node View descendants register their portals.

That might make it less necessary to put "key" in quotation marks. We're just talking about keys of a map.

We might also not want to call this a tree since this data structure itself is not one. Any other ideas?

src/contexts/PortalTreeContext.ts
```diff
@@ -0,0 +1,22 @@
...
+  null as unknown as Record<string, ReactPortal[]>
+);
+
+/**
+ * Node views that don't have any React node view ancestors
+ * can specify their parent node view as the "root", and
+ * will be have their portals rendered as direct children of
+ * the ProseMirror component.
+ */
+export const PORTAL_TREE_ROOT_KEY = "root";
```
tilgovi, line 22: Would this be an appropriate use of `Symbol`?

SMores: Yeah, seems like it!

src/contexts/PortalTreeContext.ts
SMores: Yeah I'm gonna rename this "PortalRegistryContext", I think. I don't know that inverting the subject affects the readability much; it's just hard to explain this super concisely (but also this documentation is entirely internal; this isn't exported anywhere).

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -207,10 +216,29 @@ export function createReactNodeViewConstructor(
       />
     );
 
+    if (contentDOM) {
+      // Store this node view's registration key on the contentDOM
+      // so it can be retrieved by its children.
+      contentDOM.dataset["registrationkey"] = key;
```
tilgovi, line 222: Maybe we can make it clearer what this is and say `react-pm-key` or something?

tilgovi: It feels like this key in some sense identifies this node, which makes me tempted to place it on `dom` instead of `contentDOM` even if that means descendants will have to traverse a little bit farther to find it. Does that make any sense?

tilgovi: Maybe we should also assign this constant value to a symbol.

tilgovi: I guess I mean to say "assign a symbol to this constant value".

tilgovi: And by that I don't mean `Symbol`, just a variable. 😆

src/contexts/PortalTreeContext.ts
tilgovi: I guess if this is going to be put in a DOM node's dataset it might need to be a string.

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: Yes, that's great

src/contexts/PortalTreeContext.ts
tilgovi: > A map of portal hosts to portals.
>
> React Node Views create entries in this map under unique keys, and register themselves in the values of the keys belonging to their nearest React Node View ancestors.

How's that?

src/contexts/PortalTreeContext.ts
SMores: It's not, though! This is the fallback for when we couldn't find a DOM node ancestor

src/contexts/PortalTreeContext.ts
tilgovi: Edited again.

src/contexts/PortalTreeContext.ts
tilgovi: Ah, yes. Cool.

src/contexts/PortalTreeContext.ts
SMores: > register themselves in the value for the key belonging to their nearest React Node View ancestor.

This was challenging for me to parse

SMores: /**
 * A map of node view keys to portals.
 *
 * Each node view registers a portal with under its parent's
 * key. Each can then retrieve the list of portals under their
 * key, allowing portals to be rendered with the appropriate
 * hierarchy.
 */

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: (I'm gonna leave it on contentDOM, though; if that's what ProseMirror says "is the DOM at that position", who am I to argue?)

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -152,9 +160,10 @@ export function createReactNodeViewConstructor(
           decorations={decorations}
           isSelected={isSelected}
         >
-          {ContentDOMElementType && (
-            <ContentDOMElementType
-              // @ts-expect-error There are too many HTML tags, so typescript won't compute this union type
+          {childPortals}
+          {contentDOM && (
+            <div
+              style={{ display: "contents" }}
```
tilgovi, line 166: Can we leave this change out of this PR and discuss it separately? I know you said outside GitHub that this might result in React warnings in some situations where the DOM hierarchy is otherwise invalid.

src/contexts/PortalTreeContext.ts
tilgovi: Something like that seems fine.

src/nodeViews/createReactNodeViewConstructor.tsx
tilgovi: ProseMirror tells us the starting point for our search, but that need not be the same place as where we stash the key. I'm not sure we're arguing with ProseMirror at all here.

tilgovi: I might be overthinking this, though, and certainly this is an internal detail we can change later. I think I am reaching out with my imagination towards a need to identify the nodes in the DOM for other reasons and feeling like the identity of a node is more in line with its DOM than its content DOM.

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: Yeah; I'm going to go back to "use a span if contentDOM is inline otherwise use a div", then, I guess. The current behavior is simply wrong for lists (and the HTML5 spec does say that the only valid children of a `ul` is 0 or more `li`s, even if Firefox does actually render it correctly)

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -106,9 +117,19 @@ export function createReactNodeViewConstructor(
...
+    // Use a span if the provided contentDOM is in the "phrasing" content
+    // category. Otherwise use a div. This is our best attempt at not
+    // breaking the intended content model, for now.
+    //
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content
+    const ContentDOMWrapper =
+      contentDOM &&
+      (phrasingContentTags.includes(contentDOM.tagName.toLocaleLowerCase())
+        ? "span"
+        : "div");
```
tilgovi, line 129: 😆 I typed out a suggestion to do something like this and then I deleted it because I decided to leave that for a discussion of a separate ticket.

src/contexts/PortalRegistryContext.ts
```diff
@@ -0,0 +1,23 @@
+import { ReactPortal, createContext } from "react";
+
+/**
+ * A map of node view keys to portals.
+ *
+ * Each node view registers a portal with under its parent's
```
tilgovi, line 6: Check "with under"

src/hooks/useNodeViewPortals.tsx
```diff
@@ -0,0 +1,71 @@
...
+} from "../contexts/PortalRegistryContext.js";
+import { RegisterElement } from "../nodeViews/createReactNodeViewConstructor.js";
+
+import { useEditorState } from "./useEditorState.js";
+
+type Props = {
+  portals: Record<PortalRegistryKey, ReactPortal[]>;
+};
+
+function PortalTree({ portals }: Props) {
```
tilgovi, line 17: Should we call this `PortalRegistry` in light of the context it provides being `PortalRegistryContext`?

SMores: I think my answer is no (but there's still probably some other better name for it). Here's why:

This component actually renders the entire portal tree. The fact that there's a context and a registry at all is an implementation detail; the only components that consume the "portal registry" are descendants of this component. 

This component directly renders the "top-level" portals. Each of them checks the registry and then render their direct child node view portals, and so on and so forth until the whole tree has been rendered.

No other components ever actually need to be aware of or consume the portal registry, so I think it makes more sense to think of this component as `NodeViews` or `Portals`. `PortalTree` is maybe somewhat redundant, since all React components are trees, more-or-less, so maybe we should call it `NodeViewPortals` or something instead?

tilgovi: Or even just `NodeViews`?

src/hooks/useNodeViewPortals.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const editorState = useEditorState();
+  const rootPortals = editorState && portals[PORTAL_REGISTRY_ROOT_KEY];
```
tilgovi, line 19: Why do we care about the editor state here?

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -207,10 +231,27 @@ export function createReactNodeViewConstructor(
...
 
+    if (contentDOM) {
+      // Store this node view's registration key on the contentDOM
+      // so it can be retrieved by its children.
+      contentDOM.dataset[REACT_PM_KEY] = key;
+    }
+
     const unregisterElement = registerElement(
+      // We register the element with the nearest ancestor's registration key
+      () => {
```
tilgovi, line 242: If there's a reason this is a callback rather than just as tring argument, can you please remind me?

tilgovi: I think we talked about something with the `editorView` needing to be done updating to be sure that `domAtPos` can find an element at the given position. I'm just surprised because I didn't understand function state setters in React to be anything but synchronously invoked.

tilgovi: If that's it, I wonder if we should have our own little microtask in this file that we ensure is queued that can register all unregistered node views within a batch update.

tilgovi: But that's maybe kinda what's happening here, but via React's state setting machinery, maybe?

tilgovi: Like we're gonna call setters one or more times and React will take care of scheduling an update as a result.

tilgovi: Except we're already inside a layout effect creation at this point, so I think each state set might be synchronous, which is kinda why I wonder if it'd be better to explicitly batch them.

src/hooks/useNodeViewPortals.tsx
SMores: Oh, we don't anymore! Haha good catch

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: Yeah, it's because when we call `registerElement`, ProseMirror is still descending into the node view tree and hasn't assigned view.docView yet, so this call to domAtPos would fail. Not my favorite.

src/hooks/useNodeViewPortals.tsx
SMores: Yup, I like `NodeViews`!

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: Ah sorry GitHub hadn't shown me this whole thread when I responded. I think I'm into the idea of explicitly batching/queuing this work, but I'm not sure I have a clear picture of what that would actually look like as you're describing it!

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -207,11 +241,28 @@ export function createReactNodeViewConstructor(
...
+    queueMicrotask(() => {
+      batch(() => {
+        unregisterElement = registerElement(
+          findNearestRegistryKey(editorView, getPos()),
+          element,
+          dom as HTMLElement,
+          key
+        );
+      });
+    });
```
SMores, line 265: @tilgovi I feel like you were describing something more clever than this, but... this _will_ work, right? We'll queue one microtask per node view constructor, and they'll all run after the current task queue? Are there performance implications for queuing one microtask per node view constructor rather than one microtask for all of them?

tilgovi: > We'll queue one microtask per node view constructor, and they'll all run after the current task queue?

Yes.

> Are there performance implications for queuing one microtask per node view constructor rather than one microtask for all of them?

Yes, both in function overhead and because we're calling `batch` around individual registrations rather than a group of them. I was kind hoping we could batch all of the outstanding registrations.

tilgovi: Is this getting too complicated? I'm nervous. 😆

tilgovi: Ooooh, wait, if we're not already can't we move this batch call to be around the call to update the editor view and then not worry about this here?

tilgovi: Orrr..... are we already and is that _why_ `setState` doesn't actually run its function immediately? 🤯

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -207,11 +273,31 @@ export function createReactNodeViewConstructor(
...
+    // ProseMirror has finished constructing the node view descriptor
+    // tree.
+    ensureMicrotask(() => {
+      unregisterElement = registerElement(
+        findNearestRegistryKey(editorView, getPos()),
+        element,
+        dom as HTMLElement,
+        key
+      );
+    });
```
tilgovi, line 300: This is great. Do we have to worry about the nodeview getting destroyed before this runs? I'm not sure it's possible, given that we're in a batch context right now, but it honestly hurts my brain and I'm sorry.

SMores: Hahahaha I was wondering the same... It doesn't seem possible, right? Like, presumably by the time this runs, all of the state updates have "settled" (because batch context), so any destruction couldn't be triggered until the _next_ event loop. At least, that's what it feels like. It should be pretty obvious if this does break in that way (if "unregisterElement is not callable", this is why), though, so maybe it's ok to assume for now and account for it if it's possible to mess up? Just anecdotally, I didn't have any issues testing it with the demo app.

SMores: I'm not sure I would know what to do if it destroy was called before the microtask runs. I guess we would maybe want to "cancel" the microtask, which could look like removing the task from the taskQueue? Maybe that's not so bad, actually

tilgovi: Yeah, exactly.

tilgovi:
Alternate wild idea instead of the microtask queue: have inner nodes use a mutation observer to watch the direct children of their content dom and claim their portals from the main portal registry?

SMores:
> Alternate wild idea instead of the microtask queue: have inner nodes use a mutation observer to watch the direct children of their content dom and claim their portals from the main portal registry?

Oh.... hm. Maybe this is exactly what you're saying; it's occurring to me that we could:

- Store a simple map from registration key to element
- Store a list of child keys in state in each node view component (initially empty)
- Place the registration key on the _`dom`_
- Use a mutation observer to watch for children added to `contentDOM`
- When a child has been added to `contentDOM`, read its key, and update the component state to include that key

Though... I think this would make it tricky (maybe impossible?) to support situations with "gaps" in the node view component tree, like if there's a ReactNodeView(grandparent) -> toDOM(parent) -> ReactNodeView(child) dealio.

I went on a whole ride there but I came out thinking that I still prefer what we have! Haha

tilgovi:
You're right about gaps being an issue with that approach. You can listen deeply with mutation observers, but I wouldn't want every level of the tree listening to every level below it. That sounds like it could get gnarly.

src/hooks/useNodeViewPortals.tsx
```diff
@@ -0,0 +1,64 @@
...
+              (p) => p !== portal
+            ),
+          };
+        });
+      };
+    },
+    []
+  );
+
+  return { portals: <NodeViews portals={portals} />, registerPortal };
```
tilgovi, line 63: Should we rename this key? It's not a list of portals, but a single `NodeViews` element now, yeah?

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -67,6 +76,58 @@ export type ReactNodeViewConstructor = (
   ...args: Parameters<NodeViewConstructor>
 ) => ReactNodeView;
 
+const REACT_PM_KEY = "reactpmkey";
+
+function findNearestRegistryKey(editorView: EditorView, pos: number) {
+  let parentElement: HTMLElement | null = editorView.domAtPos(pos, 0)
```
tilgovi, line 82: Can we use `closest(selector)` here with a `[data-react-pm-key]` or something?

tilgovi: I don't know if that requires you to actually add things as `data-` attributes instead of in the underlying `dataset`, but it might be nice even if you do have to do that!

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -67,6 +76,58 @@ export type ReactNodeViewConstructor = (
...
+}
+
+/** A queue of registration tasks to be executed */
+const taskQueue: Array<() => void> = [];
+
+/**
+ * Flush the task queue, executing all tasks
+ */
+function flushTaskQueue() {
+  const { length } = taskQueue;
```
tilgovi, line 98: Should we just splice and then iterate over the elements of the result rather than removing them one-by-one?

tilgovi: As long as we know that registration can't trigger cancellation. 🤯

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -67,6 +76,58 @@ export type ReactNodeViewConstructor = (
...
+  for (let i = 0; i < length; i++) {
```
tilgovi, line 99: We might need this to be `while(taskQueue.length)` if `ensureMicrotask` is doing `if (!taskQueue.length)`.

Again, not sure it's possible to have any race conditions here, but I'm trying to be cautious.

SMores: Oh, that's fine, I like that better

src/nodeViews/createReactNodeViewConstructor.tsx
tilgovi: Looks like `dataset` is a read-write proxy for setting `data-` attributes, so you should be able to use an attribute query to find the closest ancestor using `parentElement.closest("[data-reactpmkey]")`.

tilgovi: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

SMores: Oh, right! Awesome, great thought!

src/hooks/useNodeViewPortals.tsx
SMores: Yeah, I just took all of this and collapsed it into `useNodeViews` to avoid figuring out what to call this :) haha

src/hooks/useNodeViews.tsx
```diff
@@ -0,0 +1,75 @@
...
+        return {
+          ...oldPortals,
+          [registrationKey]: newChildPortals,
+        };
+      });
+
+      return () => {
+        setPortals((oldPortals) => {
+          const oldChildPortals = oldPortals[registrationKey] ?? [];
+          const newChildPortals = oldChildPortals.filter((p) => p !== portal);
```
tilgovi, line 51: Should we care about omitting this from the result if it's now zero-length?

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -67,6 +76,58 @@ export type ReactNodeViewConstructor = (
   ...args: Parameters<NodeViewConstructor>
 ) => ReactNodeView;
 
+const REACT_PM_KEY = "reactpmkey";
+
+function findNearestRegistryKey(editorView: EditorView, pos: number) {
+  const parentElement = editorView.domAtPos(pos, 0).node as HTMLElement;
+  const nearestElementWithKey = parentElement?.closest(
+    `[data-${REACT_PM_KEY}]`
+  ) as HTMLElement | null;
```
tilgovi, line 85: What is this if you don't cast it?

tilgovi: Oh, maybe `Element`?

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -67,6 +76,58 @@ export type ReactNodeViewConstructor = (
...
+}
+
+/**
+ * Cancels a queued microtask by removing it from the
+ * task queue.
+ */
+function cancelMicrotask(microtask: () => void) {
+  taskQueue.splice(taskQueue.indexOf(microtask), 1);
+}
+
```
tilgovi, line 130: I'm having second thoughts about this task queue thing and wondering if we're duplicating the work of the React scheduler handling the functional state setter calls.

There's something nice about explicitly deferring the work, but I wish we could control the flush timing better than `queueMicrotask`. Ideally right after the editor view update, inside the same layout effect, I think?

I get that this is hard because the node view constructors are bound to a register function that's defined higher in the component hierarchy. To do what I'm thinking we might need to change `<ProseMirror>` props and handle `useNodeViews` internally. That might actually be awesome, though.

If you're not in a hurry, let's sit with this for a moment. Do you mind if we take a look at what it would look like to push `useNodeViews` down inside `<ProseMirror>` (inner, maybe) such that make we could maybe do some of this in a less complicated way?

tilgovi: I see the work you've put in to keep up with the evolving thinking here and very much appreciate you already having done a lot of work to render our thoughts! But if we can keep holding this as a draft for a moment, I think we might be able to do better!

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: Yeah, domAtPos returns a `Node` and closest returns an `Element`, neither of which have the APIs we need here. We could use type guards instead of casting, but... I'm pretty confident that these will _always_ be HTMLElements

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: I feel ya. It doesn't even feel right to have to rely on the view descriptor tree at all, since ProseMirror already has the node tree fully composed when the node view constructors are being called! I can't say for sure there's a way to do this without relying on the node view descriptor tree, but if there is, I'd love to find it. I've just published this branch as-is as a pre-release version (`0..2.3-0`), so we can keep working on this branch until we feel good about it

tilgovi: Thanks!

src/hooks/useNodeViews.tsx
SMores: Hm. I suppose it is a memory leak opportunity. We can do that!

src/nodeViews/createReactNodeViewConstructor.tsx
tilgovi: Alternatively, we could actually check `instanceof HTMLElement` and fall back to the root key in the event that `domAtPos` is not an `HTMLElement` or we can't find a closest ancestor with that property that is?

I've wondered what would happen if you tried to put an `<svg>` node in a ProseMirror document and have a `<text>` descendant of it. Would it be editable? These would be `SVGElement`. Interestingly, `SVGElement` also has `dataset`.

PR #26, opened 2023-04-28
RFC: Hide node view wrappers from CSSOM with display: contents

smoores-dev:
A pain point with the current React Node View implementation is that components are "wrapped" with several layers of DOM that exist solely for the purpose of providing ProseMirror with "handles" synchronously in node view constructors. Each node view component ends up wrapped with one element (the node view `dom`), and its children are wrapped with _two_ elements (the node view `contentDOM`, _and_ the wrapper that we render in React so that we know where to put the `contentDOM` node).

Because these nodes affect content models and styling, it was important to expose them to users. However, this PR removes that configuration point, instead leaving users with a simple API that only requires a component. Under the hood, we _always_ create a `div` or `span` (details below) for the `dom` (and `contentDOM`/`contentDOM` wrapper for non-leaf nodes), and we always set the `display` style to `contents`. This means that the [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) treats the wrappers as if the weren't there (more or less), allowing users to essentially ignore them.

One note of significance is that this can still produce invalid _DOM_, even if the resulting CSSOM (after the `display: contents` has been applied) is valid. To attempt to mitigate this, we use `span`s in places where we expect elements to be used inline:

1. If `node.isInline`, its `dom` will be a `span`
2. If `node.textBlock`, its `contentDOMWrapper` and `contentDOM` will be `span`s

Another important consideration is that there are still some cases where users have to be aware of these wrappers to some extent. Specifically, this change enables writing CSS like this:

```css
ul li {
  display: flex;
}
```

because flexbox will ignore the `display: contents` contentDOMWrapper and contentDOM elements. On the other hand, this CSS will not work:

```css
ul > li {
  color: blue;
}
```

because query selectors still operate on the DOM, and don't ignore `display: contents` elements.

### DIFF
--- a/src/nodeViews/createReactNodeViewConstructor.tsx
+++ b/src/nodeViews/createReactNodeViewConstructor.tsx
@@ -1,7 +1,6 @@
 import type { Node } from "prosemirror-model";
 import type {
   Decoration,
-  DecorationSource,
   EditorView,
   NodeView,
   NodeViewConstructor,
@@ -15,8 +14,6 @@ import React, {
 } from "react";
 import type { ComponentType, ReactNode } from "react";
 import { createPortal } from "react-dom";
-
-import { phrasingContentTags } from "./phrasingContentTags.js";
 
 export interface NodeViewComponentProps {
   decorations: readonly Decoration[];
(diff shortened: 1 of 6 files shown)


src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -85,38 +82,36 @@ export type ReactNodeViewConstructor = (
...
-      innerDecorations
-    );
-
     let componentRef: NodeViewWrapperRef | null = null;
 
-    const { dom, contentDOM, component: ReactComponent } = reactNodeView;
+    const domElementType = node.isInline ? "span" : "div";
+    const dom = document.createElement(domElementType);
+    dom.dataset["pmnodetype"] = node.type.name;
+    dom.dataset["pmdom"] = "true";
```
SMores, line 99: I added these for two reasons:

1. They make the DOM a little clearer when you inspect it; it's much easier to understand what all these divs and spans are doing when they have these data attributes
2. If someone _really_ needs to for some reason, they can select these with CSS or Javascript. There are still queries that would be very challenging to make (like selecting the `dom` wrapper for a node with a specific attribute value), but I think this at least narrows down those edge cases considerably.

smoores-dev:
I think that this may not actually be safe! In Firefox at least, `display: contents` elements return 0-sized rectangles for `getClientBoundingRect`, which breaks `view.coordsAtPos` etc.

tilgovi:
I think we close this in favor of the other directions you've been exploring!

PR #27, opened 2023-05-04
Use a Plugin to assist in nesting node views according to node hierarchy

smoores-dev:
Alternative approach to #25. Instead of relying on stashing keys on dom nodes (which required queuing microtasks, since the dom reconciliation is still in progress during nodeview construction), this approach adds a ProseMirror Plugin that associates keys with nodes via their positions, and tracks node positions across transaction applications. The map from position to key allows node views to consistently find the key of their nearest React node view ancestor, since all node views have access to their own node's position.

This approach also adds sorting by position for child portals, so child element order is stable in the React tree and matches the order in the document.

Because this uses a Plugin with a `state` key, it can't be added by the `ProseMirror` component (the way that the `componentEventListenerPlugin` is, for example). Instead, this plugin is exported, and must be added by consumers. Everything will still work if consumers don't add this plugin; in fact, this change still an improvement even without the plugin, because of the stable sorting of React elements. But I've also updated the documentation to be _very clear_ (and repetitive) about the need to use this plugin alongside `useNodeViews`.


src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -106,9 +122,21 @@ export function createReactNodeViewConstructor(
...
+    // Use a span if the provided contentDOM is in the "phrasing" content
+    // category. Otherwise use a div. This is our best attempt at not
+    // breaking the intended content model, for now.
+    //
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content
+    const ContentDOMWrapper =
+      contentDOM &&
+      (phrasingContentTags.includes(contentDOM.tagName.toLocaleLowerCase())
+        ? "span"
+        : "div");
```
tilgovi, line 134: We can maybe just yank this out and land it as a simple PR first.

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -130,6 +158,16 @@ export function createReactNodeViewConstructor(
...
 
+      const portalRegistry = useContext(PortalRegistryContext);
+      const childRegisteredPortals = portalRegistry[key];
+      const childPortals = useMemo(
+        () =>
+          childRegisteredPortals
+            ?.sort((a, b) => a.getPos() - b.getPos())
+            .map(({ portal }) => portal),
+        [childRegisteredPortals]
+      );
```
tilgovi, line 169: Can we maintain them in sorted order in the registry to make this easier?

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -250,5 +297,5 @@ export function createReactNodeViewConstructor(
     };
   }
 
-  return nodeViewConstructor;
+  return Object.assign(nodeViewConstructor, { [REACT_NODE_VIEW]: true });
```
tilgovi, line 300: super nit: just writing `nodeViewConstructor[REACT_NODE_VIEW] = true` might be easier to understand.

src/plugins/componentEventListenersPlugin.ts
```diff
@@ -0,0 +1,41 @@
+import { Plugin, PluginKey } from "prosemirror-state";
```
tilgovi, line 1: If you want to make this diff smaller you could do this reorganization first, even if this will be the only plugin for a moment.

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: Yeah, Typescript just likes this better haha

src/hooks/useNodeViews.tsx
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+  const reactNodeViews = useMemo(() => {
+    const nodeViewEntries = Object.entries(nodeViews);
+    const reactNodeViewEntries = nodeViewEntries.map(([name, constructor]) => [
+      name,
+      createReactNodeViewConstructor(constructor, registerPortal),
+    ]);
+    return Object.fromEntries(reactNodeViewEntries);
+  }, [nodeViews, registerPortal]);
```
tilgovi, line 76: I know we talked a lot outside of GitHub about how to make sure we re-register when nodes get new keys, but I can't remember if we've resolved it.

Can we require the state (editor or plugin) as an argument to this hook and use the registry seed as a dependency to force the recreation of the node views?

src/nodeViews/createReactNodeViewConstructor.tsx
tilgovi: I figured. 😆

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: I _think_ the answer is no; we want to sort them when we render, not when we register, because the positions could change, right? Like if you swapped two nodes, ProseMirror might keep the node views around, so we wouldn't re-register, but the parent node view would get updated, so it would have a chance to check the new positions and put them in the right order

src/hooks/useNodeViews.tsx
SMores: As it is, it works correctly. When a node view is updated, if its key has changed, it will return false from the update method, forcing it and its children to be recreated with the new keys. This lets us incrementally refresh keys as nodes are changed, which seems better than forcing ProseMirror to rebuild all of the node views every time the editor state is reinitialized. If we wanted to instead force rebuild all of the node views, I think what you're suggesting would work, though!

src/nodeViews/createReactNodeViewConstructor.tsx
tilgovi: I'm forgetting that the registrations here are for _descendants_, not just children, in the event that there are non-React node views between here and those descendants.

Is there something we could do so that there's a React component for every node view, even if it's a non-React node view, and then the React component tree would reflect the document easily? In `NodeViews` could we iterate over the children and render either their registered portals or a stub component that just does the same for its children? That might allow node views to not even register with an ancestor, but just register a portal for their key.

src/hooks/useNodeViews.tsx
tilgovi: Ah, right, I forgot that we had originally discussed trying to keep track of the host ancestor's key, but eventually settled on nodes keeping track of their own key. Assuming the document changes at all, the top-level nodes have a chance to check their keys and force a redraw for themselves (and their descendants) if it changed.

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: This _seems_ like it should be possible, but I'm finding it challenging in practice. I can construct the initial tree (which is neat), but:

I started out trying to simply use `useEditorState` to retrieve the positionRegistry and look up the children of the current node:

```tsx
function useChildNodeViews(node: Node, pos: number) {
  const portals = useContext(PortalRegistryContext);

  const editorState = useEditorState();
  if (!editorState) return null;

  const positionRegistry = reactNodeViewPlugin.getState(editorState)!;
  console.log(positionRegistry);

  const childPositions: { node: Node; pos: number }[] = [];
  node.forEach((child, offset) => {
    childPositions.push({ node: child, pos: pos + 1 + offset });
  });

  return childPositions.map(({ node, pos }) => {
    const key = positionRegistry.get(pos)!;
    const portal = portals[key];
    return portal ?? <PlainNodeView key={key} node={node} pos={pos} />;
  });
}
```

This has a gnarly state tearing issue: the `getPos`  function passed to node views is based state in the view descriptors, so it's "stale" at render time compared to the plugin state returned by `useEditorState`. The result is that you get into infinite loops whenever you insert a character before an existing node; the node view keeps finding its _own portal_ in the registry, because it's using it's old position to look up its first child.

This led to a second approach that feels bad and doesn't work, but I think I'm too tired to figure out why.

```tsx
function useChildNodeViews(node: Node, getPos: () => number) {
  const portals = useContext(PortalRegistryContext);
  const [children, setChildren] = useState<JSX.Element[]>([]);

  useEditorEffect(
    (view) => {
      if (!view) return;
      console.log("in usechildnodes");

      const positionRegistry = reactNodeViewPlugin.getState(view.state)!;

      const childPositions: { node: Node; pos: number }[] = [];
      node.forEach((child, offset) => {
        childPositions.push({ node: child, pos: getPos() + 1 + offset });
      });

      setChildren(
        childPositions
          .filter(({ node }) => !node.isText)
          .map(({ node, pos }) => {
            const key = positionRegistry.get(pos)!;
            // console.log(node.type.name, node.textContent, key);
            const portal = portals[key];
            return portal ?? <PlainNodeView key={key} node={node} pos={pos} />;
          })
      );
    },
    [node, portals, getPos]
  );

  return children;
}
```

This way, the position isn't looked up until the state and view have been reconciled. This is... better? But it still isn't right. Some categories of changes to the doc (e.g. trying to split a list item when there's another list item after it) are totally broken, though basic editing more or less works.

I'll push this attempt to another branch so you can take a look if you like!

SMores: https://github.com/nytimes/react-prosemirror/pull/28

tilgovi: Thanks!

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: 👍 #29

src/hooks/useEditorView.ts
```diff
@@ -58,6 +58,7 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
   props: DirectEditorProps
 ): EditorView | null {
   const [view, setView] = useState<EditorView | null>(null);
+  // const [keyRegistry];
```
tilgovi, line 61:
```suggestion
```

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -215,9 +247,11 @@ export function createReactNodeViewConstructor(
     );
 
     const unregisterElement = registerElement(
+      findNearestRegistryKey(editorView, getPos()),
+      getPos,
       element,
       dom as HTMLElement,
-      Math.floor(Math.random() * 0xffffff).toString(16)
+      key
```
tilgovi, line 254: Is there any reason to have the portal registry key and the React key be different?

tilgovi (commented):
This is really close, but I added just a couple more questions and thoughts!

src/plugins/reactNodeViewPlugin.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      const next = new Map<number, string>();
+      const nextKeys = new Set<string>();
+      newState.doc.descendants((_, pos) => {
+        const prevPos = tr.mapping.invert().map(pos);
+        const prevKey = value.get(prevPos) ?? createRegistryKey();
+        // If several consecutive nodes are added in a single transaction,
+        // their positions will map back to the same previous position.
+        // We need a new key for each new node, so we have to verify uniqueness
+        // here.
```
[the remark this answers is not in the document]
SMores: Doing this for each step doesn't help! Actually, that's a good point, even the addition of one new node is enough to trigger this.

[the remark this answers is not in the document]
tilgovi: Cool, maybe we can make the comment a bit simpler, then.

src/nodeViews/createReactNodeViewConstructor.tsx
SMores: The registration key isn't unique; all of the children of a given node will register with the same key. But they're not actually different; the react key is the key that the children of _this node_ will register with. Does that make sense? Hoping I'm answering the right question haha.

src/plugins/reactNodeViewPlugin.ts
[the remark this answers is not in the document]
SMores: Done

src/nodeViews/createReactNodeViewConstructor.tsx
tilgovi: Ahh, right. Okay. And `key` _is_ the registry key of this node. Maybe naming changes can help. We have a key for ourselves, and the key of the nearest ancestor, yeah?

Are the arguments here getting busy? Should we move `createPortal` here and then register that result?

tilgovi: I think the naming I'm referring to is that "registry key" led me to think one key was a registry key and one key was a rendering key, but both are the same "kind" of key, the difference is which element they refer to.

SMores: Agreed, the naming is not super clear! I'll try to see what I can come up with (I'll try registering just the portal, too, that makes sense)

src/plugins/__tests__/reactNodeViewPlugin.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    const initialPluginState =
+      reactNodeViewPlugin.getState(initialEditorState)!;
+
+    const nextEditorState = initialEditorState.apply(
+      initialEditorState.tr.replace(
+        0,
+        0,
+        new Slice(Fragment.from(schema.nodes.list.createAndFill()), 0, 0)
+      )
+    );
```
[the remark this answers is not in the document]
SMores: (I had messed up the schema above, ignore this!)

SMores:
Hey @tilgovi, curious about where your thinking is at on this one!

saranrapjs:
I've pulled in the changes from `main` and tweaked the documentation. Would love to figure out what's remaining before merging! (One thing on my list is poking around to make sure there aren't other regressions here)

README.md
```diff
@@ -367,7 +367,7 @@ const reactNodeViews = {
 const editorState = EditorState.create({
   schema,
   // You must add the react if you use
-  // the useNodeViews hook.
+  // the useNodeViews or useNodePos hook.
```
smoores-dev, line 385: Not strictly necessary; you can only use `useNodePos` if you're already using `useNodeViews`, since that's how you register React components as node views haha

smoores-dev:
I suspect that this is good to go; the good folks at https://dskrpt.de/ have been using this branch for months! Haha

README.md
saranrapjs: the `useNodePos` hook doesn't throw if the state field provided by this plugin isn't registered, so i feel like we want to note it somewhere? the reason i ran into the omission of this plugin in the first place was that we were using `useNodePos` and node updates were being silently eaten by ProseMirror because the node view position was off by way of that hook falling back to `0` when the `react()` plugin isn't found in state...

is there a different way you'd articulate the relationship between `useNodePos` and `react()`? or are you saying that as of this branch, using `useNodeViews` without `react()` won't work at all? if that's the case i wonder if we should take a harder line and throw (or something) when we can't find the plugin

smoores-dev: It won't throw, but it won't work properly (in that, it won't organize the React hierarchy correctly). I wouldn't be opposed to throwing if we can't find the plugin, though, I think that's probably a good idea actually! It should be pretty straightforward to do, since we're defining/have access to the plugin key!

All I was pointing out in my comment is that "using `useNodeViews`" already covers "using `useNodePos`" haha.

saranrapjs: ahhhh. i understand now!

saranrapjs: > in that, it won't organize the React hierarchy correctly

do things not render correctly? or do they just present incorrectly in React devtools?

smoores-dev: Uhhhh both I guess. Specifically, this PR makes it so that React Components of ProseMirror children are actually rendered as children of their parent nodes' React Components, which is important if you want context to flow correctly from parent to child. The issue that we were solving at dskrpt was for a collapsible detail/summary node, where we wanted to use context to pass parent state to the child.

So the DOM is rendered correctly either way, but since React hierarchy matters for context flow, things still don't work properly without the plugin.

src/hooks/useNodePos.tsx
```diff
@@ -15,8 +15,9 @@ export function NodePosProvider({ nodeKey, children }: Props) {
   const editorState = useEditorState();
   if (!editorState) return <>{children}</>;
   const pluginState = reactPluginKey.getState(editorState);
+  if (!pluginState) throw new Error('Can\'t find the react() ProseMirror plugin. Was it added to the EditorState.plugins?')
```
smoores-dev: How would you feel about moving this check [here](https://github.com/nytimes/react-prosemirror/blob/main/src/nodeViews/createReactNodeViewConstructor.tsx#L122-L124)? That would cover this case as well, and reduce the likelihood that someone runs into unexpected behavior re: context flow.

saranrapjs (approved):

smoores-dev:
Haha unfortunately @saranrapjs you're not a codeowner, and I opened the PR, sooo we need someone else to approve :P

saranrapjs:
I think I _am_ now a code owner, but we have "Require approval of the most recent reviewable push" enabled (is why my approval doesn't count, I think). Will check in w/ some other folks in the morning!

PR #30, opened 2023-05-25
Replace getPos with useNodePos

smoores-dev:
The `getPos` function passed to Node Views from ProseMirror is unsafe to use in the render function of React Components, because its return value is derived from the state of the EditorView.

Instead, since #27 implements a plugin tracking the position of each node, this PR exposes a `useNodePos` hook that, under the hood, retrieves the current position of the node in the editorState via that node's key.

### DIFF
--- a/.yarn/versions/bf34ddf6.yml
+++ b/.yarn/versions/bf34ddf6.yml
@@ -1,0 +1,2 @@
+releases:
+  "@nytimes/react-prosemirror": minor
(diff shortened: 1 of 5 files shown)


tilgovi (commented):
Looks good, but I think we need a documentation update for `useNodeViews` that removes the prop, no?

smoores-dev:
@tilgovi I don't think so, no! `getPos` is still passed to the node view constructor (and it's safe to use there), it's just not passed to the react component as a prop. So that function signature hasn't changed.

tilgovi (approved):

PR #33, opened 2023-05-31
Implement a plugin that tracks node positions.

SMores:
The plugin tracks a unique key for each (non-text) node in the document, identified by its current position. This provides a mechanism for node views to uniquely identify themselves via a key, as well as determine the position of other node views from their respective keys. Keys remain mostly stable across changes to the document.

This will enable other improvements, like a safer mechanism for accessing node position from within NodeViews and proper React Context hierarchies across node view components.


SMores:
This change is pre-work for #27, #30, and potentially #32!

SMores:
I'm fine with the factory functions, but I'm a little worried about the idea of just exporting something called `reactNodeViews`! It just seems confusing

tilgovi:
I'm curious whether the relatively close naming of `reactNodeViews` (the plugin) and `useNodeViews` (the hook) is an unfortunate near collision or actually suggests something insightful about maybe the hook being the wrong API.

SMores:
It's kind of interesting... the `reactNodeViews` plugin is mostly named as such because, opaquely, to users, you need to use it in order to use React-based node views. But it doesn't intrinsically have anything to do with node views, just provides a convenient way to track nodes across transactions.

tilgovi:
I'm kinda okay with it being less descriptive, I think. Contextually, it'll make sense, and especially with API documentation and examples. Folks will import it and include it without thinking much of it, I think.

tilgovi:
Yeah, I guess we probably do _not_ want to track state of the node views within the `reactNodeViews` plugin. So, maybe we need another name for it, and that'll help resolve the suffix issue, too.

tilgovi:
> Taking that a little further, then: What if we just named it react? That feels more inline with ProseMirror's first-party plugins, and is also how, say, Vite does its plugin naming

I feel good about that. Also, it feels like the key (:smirk:) thing that integrates React with ProseMirror is keying of nodes. That's what bridges the gap between React and ProseMirror reconciliation.

tilgovi:
Although, will that feel really awful if you import it in a file where you also import `React`?

tilgovi:
`nodeKeys()`?

SMores:
> Although, will that feel really awful if you import it in a file where you also import React?

I think it won't? I feel like, since `React` is always and must be imported specifically as "React", the big "R" is really part of it. I think it would be fine to have both.

> `nodeKeys()`?

I don't like this because explaining why it's called this requires explaining a bunch of internal details that consumers _definitely_ do not need to know. So either it's just gibberish by design or it's way too in-the-weeds.

tilgovi:
I'm fine to try `react`.

SMores:
Done!

src/plugins/__tests__/reactNodeViewPlugin.test.ts
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
tilgovi, line 1: This file needs a rename, yeah?

src/hooks/useComponentEventListenersPlugin.tsx
```diff
@@ -1,46 +1,10 @@
-import { Plugin, PluginKey } from "prosemirror-state";
```
tilgovi, line 1: Should we drop plugin from the name of this file and hook, too?

src/plugins/__tests__/reactNodeViewPlugin.test.ts
SMores: Whoops, yes, thanks

src/plugins/componentEventListenersPlugin.ts
```diff
@@ -0,0 +1,41 @@
+import { Plugin, PluginKey } from "prosemirror-state";
```
tilgovi, line 1: We can drop `Plugin` here, too, right?

src/hooks/useComponentEventListenersPlugin.tsx
SMores: Yeah!

SMores:
Cleaned those up, thanks for spotting haha

tilgovi (commented):
The plugins are meant to be public, yeah? Do we need to create an index for the plugins directory and re-export the plugins, then re-export them from the root?

tilgovi (commented):
The plugins are meant to be public, yeah? Do we need to create an index for the plugins directory and re-export the plugins, then re-export them from the root?

tilgovi:
Do we need any README changes?

SMores:
> The plugins are meant to be public, yeah? Do we need to create an index for the plugins directory and re-export the plugins, then re-export them from the root?

The `componentEventListeners` plugin is private; it gets automatically added to the EditorView by the ProseMirror component.

The `react` plugin will be public but I was thinking maybe we'd expose it in a later change when we actually use it for something (and add the docs then, as well)? Right now, adding it doesn't actually do anything for users

tilgovi (approved):

src/plugins/react.ts
```diff
@@ -0,0 +1,83 @@
...
+ * key for a given node can be accessed by that node's
+ * current position in the document, and vice versa.
+ */
+export function react() {
+  return new Plugin({
+    key: reactPluginKey,
+    state: {
+      init(_, state) {
+        const next = {
+          posToKey: new Map<number, string>(),
```
halfghaninne, line 36: Is there a reason that we are using a `string` on line 36 and `NodeKey` on the following line?

SMores: Oh interesting, yeah this is worth thinking about a bit more, maybe. `NodeKey` is just "`string` + `Symbol('portal rogistry root key')`", and technically we never actually put that root symbol in the plugin's bimap. The reason that `keyToPos` is typed as having `NodeKey` is just so that consumers don't have to do a special-case check to see whether they have the root symbol or a string before trying to get the position from the `keyToPos` map (that is, it's type-safe to do `pluginState.keyToPos.get(nodeKey)`, where `nodeKey` is of type `NodeKey`).

On the other hand, since we never actually put that root symbol in the bimap, technically it's correct that `pluginState.posToKey(pos)` always returns `string`, and you can use a `string` with any interface that accepts a `NodeKey` (since `NodeKey` is a subtype of `string`). So I think from a purely type-checking standpoint, this is "correct", but I also think it's somewhat unintuitive and might seem wrong to someone who's just using the API (which isn't _actually_ exposed to users, so this would just be "us in the future", but we should avoid confusing "us in the future", too!).

The only downside I can think of to typing `posToKey` as `Map<number, NodeKey>` is that you can technically take any result of `posToKey` and pass it to any function that expects a `string`, but if it was typed as `NodeKey`, Typescript wouldn't let you do that! I dunno what functions you'd be passing a node key to that weren't typed as NodeKey, so this seems pretty minor, as downsides go (and it's an internal API so it's not a big deal if we have to change it later).

Any thoughts after all that? Haha

SMores: Also we should probably change the Symbol name! That was from a previous version of this plugin, and it doesn't really make sense here

PR #34, opened 2023-06-01
Add support for uncontrolled ProseMirror component

SMores:
Adds a `defaultState` prop to support uncontrolled usage of the ProseMirror component! This PR ensures that `useEditorState` always returns an up-to-date EditorState, whether that's from the state prop or from the view state (if uncontrolled)

### DIFF
--- a/src/hooks/useEditorView.ts
+++ b/src/hooks/useEditorView.ts
@@ -1,8 +1,10 @@
-import type { Transaction } from "prosemirror-state";
+import type { EditorState, Transaction } from "prosemirror-state";
 import { EditorView } from "prosemirror-view";
 import type { DirectEditorProps } from "prosemirror-view";
 import { useLayoutEffect, useState } from "react";
 import { unstable_batchedUpdates as batch } from "react-dom";
+
+import { useForceUpdate } from "./useForceUpdate.js";
 
 function withBatchedUpdates<This, T extends unknown[]>(
   fn: (this: This, ...args: T) => void
@@ -14,12 +16,19 @@ function withBatchedUpdates<This, T extends unknown[]>(
   };
 }
 
-const defaultDispatchTransaction = function (
-  this: EditorView,
-  tr: Transaction
-) {
-  batch(() => this.updateState(this.state.apply(tr)));
-};
+function defaultDispatchTransaction(this: EditorView, tr: Transaction) {
+  this.updateState(this.state.apply(tr));
+}
+
+type EditorStateProps =
+  | {
+      state: EditorState;
+    }
+  | {
+      defaultState: EditorState;
+    };
+
+export type EditorProps = Omit<DirectEditorProps, "state"> & EditorStateProps;
 
 /**
  * Enhances editor props so transactions dispatch in a batched update.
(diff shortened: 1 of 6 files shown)


tilgovi:
Do we need a separate prop for this or is there a way we can make it sufficient just to not pass `dispatchTransaction`, similar to how one can just not pass `onChange` handlers to regular form controls.

SMores:
I think these are separate, right? For native elements at least, passing `onChange` doesn't determine controlledness; whether `value` is `undefined` does (at least, I'm pretty sure). You might still want/need to hook in to `dispatchTransaction` to track some kind of state, even if you don't want to lift the whole EditorState

tilgovi:
Your understanding is correct. I'm mistaken. React uses `defaultValue` and I had no memory of that, if I ever knew it. I think `defaultState` makes sense.

src/hooks/useEditorView.ts
```diff
@@ -14,12 +16,12 @@ function withBatchedUpdates<This, T extends unknown[]>(
...
-  tr: Transaction
-) {
-  batch(() => this.updateState(this.state.apply(tr)));
-};
+function defaultDispatchTransaction(this: EditorView, tr: Transaction) {
+  this.updateState(this.state.apply(tr));
+}
+
+export type EditorPropsWithDefaultState = Omit<DirectEditorProps, "state"> &
+  ({ defaultState: EditorState } | { state: EditorState });
```
tilgovi, line 24: Maybe we should just import `EditorProps as _EditorProps` and then export `EditorProps extends _EditorProps`. We'll have to repeat the definition of `dispatchTransaction` and `plugins`, but it might end up easier to read and the naming will be nice.

```typescript
export interface EditorProps extends _EditorProps {
  defaultState?: EditorState | undefined;
  state?: EditorState | undefined;
}
```

Actually, with TypeScripts interface definition merging, could we just redefine `EditorProps`? Or does that only allow us to add `defaultState` without overriding `state`?

SMores: Yeah I don't think we could make `state` optional that way. I like the idea of making it so that you can only provide `state` _or_ `defaultState`, too; defining both is sort of nonsensical

tilgovi: It might actually be more ergonomic for some use case if it were allowed to pass it with a value of undefined, but it's hard to say what that use case would be.

I guess I didn't realize that you could actually do the type algebra you had here and get something meaningfully different from my example interface.

tilgovi: I'm tempted to just default to a conservative approach that's similar to React, where it's technically possible to supply both: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1349b640d4d07f40aa7c1c6931f18e3fbf667f3a/types/react/index.d.ts#LL1746C16-L1746C16

SMores: Ah, so here's a hiccup: React can partly get away with those types because it's legal to pass no `defaultValue` _or_ `value` to an input. HTML inputs have their own default value constructors. But you _have_ to pass a `state` to `EditorView`, and we can't produce a default one without knowing your schema... which we would only know from the state. So we actually need to require at least one of them. I suppose we can just do this in code, rather than with types. What do you think?

tilgovi: No strong preference. What's your leaning?

SMores: Hmmmmmm. I think I want to leave it as is; if there are issues where we need to provide both props (with one undefined, or something), we can always relax the type constraints later. More challenging to tighten them!

SMores: Cleaned up the type declaration a bit, though

tilgovi (approved):

Issue #42, opened 2023-06-28
Deleting the last node view can throw errors


tilgovi:
As reported in https://github.com/nytimes/react-prosemirror/issues/40#issuecomment-1605224906, deleting the last character in the only remaining node view can throw errors. This is reproducible in the provided demo.

```
react-dom.development.js:12056 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at removeChildFromContainer (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:8473:23)
    at commitDeletionEffectsOnFiber (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17504:21)
    at recursivelyTraverseDeletionEffects (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17482:13)
    at commitDeletionEffectsOnFiber (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17573:15)
    at recursivelyTraverseDeletionEffects (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17482:13)
    at commitDeletionEffectsOnFiber (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17573:15)
    at recursivelyTraverseDeletionEffects (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17482:13)
    at commitDeletionEffectsOnFiber (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17530:17)
    at commitDeletionEffects (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17473:13)
    at recursivelyTraverseMutationEffects (http://localhost:5173/node_modules/.vite/deps/chunk-OB2TTU35.js?v=886cae67:17670:17)
flushSyncCallbacks @ react-dom.development.js:12056
(anonymous) @ react-dom.development.js:25651
```

tilgovi:
[...]

PR #47, opened 2023-07-24
Move rendering responsibility fully into React

smoores-dev:
This is a re-imagining of this project's integration with ProseMirror's EditorView. Instead of attempting to carefully coordinate between React and ProseMirror, this fully moves the responsibility of rendering the editor state to the DOM into React.

## How did this work before?

The primary work of this library is to manage the relationship between React and ProseMirror's EditorView. As laid out in [The Problem Statement](https://github.com/nytimes/react-prosemirror/tree/aacb3243d68a7ec65b42d88359b95b94435df166#the-problem) in the README, this is a fairly delicate relationship, and it's not without its tradeoffs. A short review:

- React updates in two phases (a side-effect-free state update phase, followed by a side-effect-ful render phase), while ProseMirror updates state and renders "all at once" (in that the two steps can't be separated)
  - This makes it challenging to store ProseMirror state in React; ProseMirror's state can't be updated until "phase 2", because it will execute side effects, so React will have "newer" state than ProseMirror during phase 1
  - Again, since ProseMirror renders synchronously after a state update, and React does not (at least, it's not guaranteed to), it is challenging to use React components to render ProseMirror node views, because ProseMirror asks for the rendered DOM from its node view constructors synchronously, and React can't provide that.

Until now, this library has relied on the default behavior of ProseMirror's EditorView for rendering non-node-view parts of the ProseMirror document, and has attempted to mitigate the challenges stated above by:

- Carefully coordinating state update timing so that we don't violate React's expectations about side effects
- Limiting access to the EditorView to times where we can be confident that its state is up to date with React's
- Wrapping all React-based node views with synchronously rendered DOM elements that can be handed to ProseMirror's EditorView during its one-phase update cycle.

## Why are we interested in alternative solutions?

The current implementation is quite good at what it does, and it's proven to be very valuable. However, there are some downsides:

- Wrapping all React-based node views is unfortunate. 
  - It can make it challenging to produce valid DOM in some scenarios. For example: If using React-based node views to render a `ul` with child `li`s, the actual `ul` and `li` elements have to be managed outside of React, in the constructor, and set as the contentDOM and DOM respectively.
  - It can make styling challenging, especially because one of the wrapping elements (the contentDOM wrapper) is produced internally to react-prosemirror and cannot be customized easily by consumers
  - It can introduce corner cases in contenteditable implementations, such as #42 (which we managed to sort of hack around [here](https://github.com/nytimes/react-prosemirror/pull/62/files#diff-b1e30e8d82f61a0adfec6889cf827bc50ee93ef0a19f0175674c0826f2243621R267-R274), thanks again to @saranrapjs for figuring this out!)
- We induce a double render for every ProseMirror state update.
  - The order of events looks like:
    - State is updated through a `dispatchTransaction` call
    - The ProseMirror component is rendered, along with all of its children
      - This includes the React-based node view components, but their state is still the previous state
    - The EditorView is updated (in a `useLayoutEffect`). This is when the node views are created/updated, but the React-based ones are not synchronously rendered.
    - React-based node views are rendered again, this time with the new state
  - This is unintuitive from a developer standpoint, particularly the fact that React-based node views can be rendered with old state, even if they were the source of the state change (see #69)

## What does this approach do differently?

This PR proposes to take over user input tracking (manual text input, at least) and rendering responsibilities from ProseMirror's EditorView, consolidating all rendering responsibilities into React.

This has several advantages:

- There is no longer any need for wrapping React-based node views in several layers of manually managed DOM elements. A React-based node view that renders only a `<p>` tag will result in only a `<p>` tag in the document.
- There is no longer any need for a double render. React-based node views will be immediately updated with any state changes, and the EditorView state will always be in sync with the React state.
- The interface for React-based node views is a little more idiomatic and hopefully easier to reason about.
- Widget decorations can also be authored with React components

The notable disadvantage to this approach is that non-React node views and widgets will have to be wrapped in a single wrapping element. This is less cumbersome than the React-based node view wrappers in the current approach, but is worth calling out.

## How does it work?

This PR extends the EditorView class and overrides its NodeView and DOMObserver properties. In their places, we provide custom NodeViews, that act exactly the same as the default implementation, but execute no-ops when their `update` methods are called, and a custom DOMObserver that only listens for selection changes, and ignores DOM mutations.

We then introduce a `beforeinput` event handler that produces ProseMirror transactions based on the events it receives, which replaces the DOMObserver functionality that we eschewed in our selection-only implementation.

We also introduce a new React component suite, one for each type of NodeView, to replace the truncated `update` methods. When state is updated, we render a `DocNodeView` component with `view.state.doc` and the document decorations. Each NodeView component is responsible for rendering its children, so this eventually renders the entire ProseMirror document (and all of its decorations) in a single React tree.


src/browser.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const ie_edge = /Edge\/(\d+)/.exec(agent);
+const ie_upto10 = /MSIE \d/.exec(agent);
+const ie_11up = /Trident\/(?:[7-9]|\d{2,})\..*rv:(\d+)/.exec(agent);
+
+export const ie = !!(ie_upto10 || ie_11up || ie_edge);
+export const ie_version = ie_upto10
+  ? (document as any).documentMode
+  : ie_11up
+  ? +ie_11up[1]!
```
tilgovi, line 13: I'd be very in favor of dropping any pre-edge IE support from this.

src/components/EditorView.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+      let element: ReactNode = Component
+        ? createElement(Component, {
+            node: childNode,
+            decorations: [],
+            // TODO: This is how I'm reading the prosemirror-view code,
+            // but I would have expected this to be broader?
+            isSelected:
+              state.selection instanceof NodeSelection &&
+              state.selection.node === node,
```
tilgovi, line 227: Yeah, I think that's right. We have used a broader definition at the Times in a custom way.

src/components/EditorView.tsx
```diff
@@ -0,0 +1,329 @@
...
+    parentElement: JSX.Element,
+    node: Node,
+    pos: number,
+    decorations: DecorationSet
+  ) {
+    const childElements: ReactNode[] = [];
+    if (node.childCount === 0 && node.isTextblock) {
+      childElements.push(<br />);
+    }
+    node.forEach((childNode, offset) => {
```
tilgovi, line 141: This this is definitely one way to do this, but if we are okay with some intermediate nodes, we may be able to let React memoize levels of this for us. The approach would be to have this be a function component that renders only one level at a time, and wrap it in `React.memo`.

But this isn't identical to how I had been doing things in my earlier attempts to use React with ProseMirror, so I'll keep reading and thinking about the call signature here and whether that works. I mention it so that you can mull it over, too.

saranrapjs:
> here we eschew prosemirror-view's EditorView entirely, in place of a custom-built React component

it doesn't look like any of the implementation here required changes to any of the existing React wrappers for ProseMirror view components, yes? is there a model where the React-based EditorView is offered as an additional import, but where we could commit to keeping compatibility with the regular ProseMirror EditorView? i can imagine scenarios where retaining that compatibility is important, presuming it doesn't really burden this library to do so

src/dom.ts
```diff
@@ -0,0 +1,48 @@
...
+export function flattenV(rect: DOMRect, left: boolean) {
+  if (rect.width == 0) return rect;
+  const x = left ? rect.left : rect.right;
+  return { top: rect.top, bottom: rect.bottom, left: x, right: x };
+}
+export function flattenH(rect: DOMRect, top: boolean) {
+  if (rect.height == 0) return rect;
+  const y = top ? rect.top : rect.bottom;
+  return { top: y, bottom: y, left: rect.left, right: rect.right };
+}
```
tilgovi, line 44: Do these add readability vis-a-vis `const x = { ...y, bottom: y.top }`?

src/dom.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function nodeSize(node: Node) {
+  return node.nodeType == 3 ? node.nodeValue!.length : node.childNodes.length;
+}
```
tilgovi, line 48: Is this potentially confusing given that ProseMirror uses `size` in some places to mean the number of document position indices the node occupies?

smoores-dev:
@saranrapjs 

> it doesn't look like any of the implementation here required changes to any of the existing React wrappers for ProseMirror view components, yes? is there a model where the React-based EditorView is offered as an additional import, but where we could commit to keeping compatibility with the regular ProseMirror EditorView? i can imagine scenarios where retaining that compatibility is important, presuming it doesn't really burden this library to do so

I think that I wouldn't be in favor of that; there's almost no overlap between the two approaches, so we'd basically be maintaining two separate libraries. What kind of compatibility requirement are you imagining, though?

src/dom.ts
smoores-dev: This entire file is directly copied from prosemirror-view's dom.ts file, with no modifications. That library does in fact alternate between discussing prosemirror nodes and dom nodes with the same vocabulary in a way that I found hard to follow. I'm not really proposing that we merge this PR as-is; I think it needs work in terms of code organization and naming, just wanted to get feedback on the approach!

tilgovi: Gotcha!

src/components/EditorView.tsx
smoores-dev: Yeah; I do think I would rather implement these functions such that they take children as arguments, rather than being responsible for cloning themselves to add children. Was just trying to get the basics down before doing too much cleanup

saranrapjs:
> What kind of compatibility requirement are you imagining, though?

@smoores-dev I'm picturing:
- Changes to non-`prosemirror-view` classes require corresponding changes in `prosemirror-view`
- Changes other libraries depend on behavior we may not have implemented to spec in our React-based `prosemirror-view`
- Bugfixes in `prosemirror-view` that introduce corner cases in our implementation

Maybe this is a question we can ask Marijn? e.g. whether there's a "safe" way we should go about maintaining an EditorView implementation that tries to stay in sync with other `prosemirror-*` libraries?

src/components/EditorView.tsx
tilgovi: Here's what I had, which was making little React components for each kind of node view, and sharing a render function between them (for some reason).

```
import isShallowEqual from "@wordpress/is-shallow-equal";
import hoistStatics from "hoist-non-react-statics";
import { DOMSerializer } from "prosemirror-model";
import type { Node as ProseMirrorNode } from "prosemirror-model";
import { EditorView } from "prosemirror-view";
import type {
  CustomNodeViewDesc,
  Decoration,
  DecorationSet,
  DirectEditorProps,
  MarkViewDesc,
  NodeViewDesc,
  NodeView as NodeViewSpec,
  ViewDesc,
  WidgetViewDesc,
} from "prosemirror-view";
import { useCallback, useLayoutEffect, useRef, useState } from "react";
import type { ComponentType, ReactNode, Ref, ReactElement } from "react";
import { createPortal } from "react-dom";

function isMarkViewDesc(viewDesc: ViewDesc): viewDesc is MarkViewDesc {
  return "mark" in viewDesc;
}

function isNodeViewDesc(viewDesc: ViewDesc): viewDesc is NodeViewDesc {
  return (
    "node" in viewDesc &&
    "nodeDOM" in viewDesc &&
    "outerDeco" in viewDesc &&
    "innerDeco" in viewDesc
  );
}

function isCustomNodeViewDesc(
  viewDesc: ViewDesc
): viewDesc is CustomNodeViewDesc {
  return isNodeViewDesc(viewDesc) && "spec" in viewDesc;
}

interface ComponentNodeViewSpec extends NodeViewSpec {
  Component: ComponentType<NodeViewDescProps>;
}

interface ComponentNodeViewDesc extends CustomNodeViewDesc {
  spec: ComponentNodeViewSpec;
}

function isComponentNodeViewDesc(
  viewDesc: ViewDesc
): viewDesc is ComponentNodeViewDesc {
  return isCustomNodeViewDesc(viewDesc) && "Component" in viewDesc.spec;
}

function isWidgetViewDesc(viewDesc: ViewDesc): viewDesc is WidgetViewDesc {
  return "widget" in viewDesc;
}

type ViewDescProps<T extends ViewDesc = ViewDesc> = {
  render: (view: EditorView, viewDesc: ViewDesc) => ReactElement | null;
  view: EditorView;
  viewDesc: T;
};

type MarkViewDescProps = ViewDescProps<MarkViewDesc>;

type NodeViewDescProps = ViewDescProps<NodeViewDesc>;

type WidgetViewDescProps = ViewDescProps<WidgetViewDesc>;

function ContentPortal(props: ViewDescProps): ReactElement | null {
  const { render, view, viewDesc } = props;

  const children = viewDesc.children.map((childViewDesc) =>
    render(view, childViewDesc)
  );

  if (viewDesc.contentDOM instanceof Element) {
    return createPortal(children, viewDesc.contentDOM);
  }

  return null;
}

const DocView = (props: NodeViewDescProps) => ContentPortal(props);
DocView.displayName = "Doc";

const MarkView = (props: MarkViewDescProps) => ContentPortal(props);
MarkView.displayName = "Mark";

const NodeView = (props: NodeViewDescProps) => ContentPortal(props);
NodeView.displayName = "Node";

const WidgetView = (props: WidgetViewDescProps) => ContentPortal(props);
WidgetView.displayName = "Widget";

export interface NodeViewProps {
  children?: ReactNode | undefined;
  node: ProseMirrorNode;
  nodeViewRef: Ref<NodeViewSpec>;
  view: EditorView;
  getPos: () => number;
  decorations: Decoration[];
  innerDecorations: DecorationSet;
}

export function nodeView(
  WrappedComponent: ComponentType<NodeViewProps>
): (node: ProseMirrorNode, view: EditorView) => NodeViewSpec {
  const wrappedComponentName =
    WrappedComponent.displayName || WrappedComponent.name || "Component";
  const displayName = `NodeView(${wrappedComponentName})`;

  const Component = (props: NodeViewDescProps) => {
    const { view, viewDesc } = props;
    const { node } = viewDesc;
    const parentViewDesc = viewDesc.parent!;
    const parentDOM = viewDesc.dom?.parentNode!;

    const getPos = useCallback(() => {
      return parentViewDesc.posBeforeChild(viewDesc);
    }, [parentViewDesc, viewDesc]);

    const ref = useCallback(
      (spec: NodeViewSpec | null) => {
        const dom = spec?.dom ?? null;
        if (dom) {
          if (
            dom instanceof HTMLElement &&
            !viewDesc.contentDOM &&
            !dom.hasAttribute("contenteditable")
          ) {
            dom.contentEditable = "false";
          }

          // @ts-expect-error
          dom.pmViewDesc = viewDesc;

          if (viewDesc.dom) {
            parentDOM.replaceChild(dom, viewDesc.dom);
          }
        } else {
          // ProseMirror has already removed the node.
          // Put it back so that React can remove it.
          // :sweat_smile:
          if (viewDesc.dom) {
            parentDOM.insertBefore(viewDesc.dom, parentDOM.firstChild);
          }
        }
        viewDesc.dom = dom;
        viewDesc.nodeDOM = dom;

        const contentDOM = spec?.contentDOM ?? null;
        if (contentDOM) {
          if (viewDesc.contentDOM) {
            Array.from(viewDesc.contentDOM.childNodes).forEach((child) =>
              contentDOM.appendChild(child)
            );
          }
        }
        viewDesc.contentDOM = contentDOM;

        if (isCustomNodeViewDesc(viewDesc)) {
          Object.assign(viewDesc.spec, spec);
        }
      },
      [parentDOM, viewDesc]
    );

    const children = ContentPortal(props);

    return (
      <WrappedComponent
        node={node}
        view={view}
        getPos={getPos}
        decorations={viewDesc.outerDeco}
        innerDecorations={viewDesc.innerDeco}
        nodeViewRef={ref}
        children={children}
      />
    );
  };

  Component.WrappedComponent = WrappedComponent;
  Component.displayName = displayName;
  hoistStatics(WrappedComponent, Component);

  return (node, view) => {
    const document = view.dom.ownerDocument;
    const domSpec = node.type.spec.toDOM?.(node);
    const { dom, contentDOM } = DOMSerializer.renderSpec(document, domSpec!);
    return { dom, contentDOM, Component };
  };
}

const keyCache = new WeakMap<ViewDesc, string>();

function keyCacheGet(viewDesc: ViewDesc): string {
  if (keyCache.has(viewDesc)) {
    return keyCache.get(viewDesc)!;
  } else {
    const key = ((Math.random() * 0xffffff) << 0).toString(16);
    keyCache.set(viewDesc, key);
    return key;
  }
}

function renderViewDesc(
  view: EditorView,
  viewDesc: ViewDesc
): ReactElement | null {
  const key = keyCacheGet(viewDesc);

  if (isComponentNodeViewDesc(viewDesc)) {
    const { Component } = viewDesc.spec;
    return (
      <Component
        render={renderViewDesc}
        view={view}
        viewDesc={viewDesc}
        key={key}
      />
    );
  }

  if (isMarkViewDesc(viewDesc)) {
    return (
      <MarkView
        render={renderViewDesc}
        view={view}
        viewDesc={viewDesc}
        key={key}
      />
    );
  }

  if (isNodeViewDesc(viewDesc)) {
    if (viewDesc.node.isText) {
      return null;
    }

    if (viewDesc.node === view.state.doc) {
      return (
        <DocView
          render={renderViewDesc}
          view={view}
          viewDesc={viewDesc}
          key={key}
        />
      );
    }

    return (
      <NodeView
        render={renderViewDesc}
        view={view}
        viewDesc={viewDesc}
        key={key}
      />
    );
  }

  if (isWidgetViewDesc(viewDesc)) {
    return (
      <WidgetView
        render={renderViewDesc}
        view={view}
        viewDesc={viewDesc}
        key={key}
      />
    );
  }

  return null;
}

export function useEditorView(props: DirectEditorProps) {
  const [mount, setMount] = useState<Node | null>(null);
  const [view, setView] = useState<EditorView | null>(null);
  const hasFocusRef = useRef<boolean>(false);

  useLayoutEffect(() => {
    if (mount && !view) {
      setView(new EditorView({ mount }, props));
    } else if (!mount && view) {
      view.destroy();
      setView(null);
    }
  }, [mount, props, view]);

  useLayoutEffect(() => {
    if (view) {
      // @ts-expect-error
      view.domObserver.start();
      if (hasFocusRef.current) view.focus();
    }
  });

  if (view) {
    const { state: oldState, ...oldProps } = view.props;
    const { state: newState, ...newProps } = props;
    if (isShallowEqual(oldProps, newProps)) {
      if (oldState !== newState) {
        view.updateState(newState);
      }
    } else {
      view.update(props);
    }

    // @ts-expect-error
    view.domObserver.stop();
    hasFocusRef.current = view.hasFocus();
  }

  // @ts-expect-error
  const children = view ? renderViewDesc(view, view.docView) : null;

  return {
    children,
    contentRef: setMount,
    view,
  };
}
```

smoores-dev:
@saranrapjs Gotcha. Yeah I suspect that when it comes down to it, it's going to be a tradeoff between the risks you've outlined here, and the risks of the approach we have now (which mostly center around the wrapping elements). We've already run into several frustrating corner cases with the wrapping elements: #42 is one of these, and #26 is an attempt to mitigate a few others. I actually haven't decided which way I personally fall on this; there are obvious downsides to both options.

smoores-dev:
@tilgovi @saranrapjs (and anyone else interesting): 

So... what do you think? Should we actually try to do this? I don't see any reason that we _literally could not_ continue implementing the rest of the EditorView API, though we might have to be thoughtful about how to handle, for example, commands that require the view (even `joinBackward` [relies on the view](https://github.com/ProseMirror/prosemirror-commands/blob/master/src/commands.ts#L15-L21), which I was kind of shocked to discover!).

I started down this road after #42; there may be a direct kludge (or even more elegant solution) that can solve that literal case, but it made me worried that there are more challenges with the general approach of wrapping elements that we initially expected.

I think the challenge with this approach, as @saranrapjs has pointed out, is figuring out how to ensure that whatever we do here actually plays nicely with existing prosemirror libraries, including user-contributed ones. In particular, this... _might_ mean some support for plugins with a view prop...? But that also seems a little too goofy for me; maybe it's acceptable to say that we only support state-only plugins, and the whole reason to use React is so that it can own the entirety of your view.

tilgovi:
Unless we want to pursue compiling a secondary ReactDOM renderer that we can flush synchronously, I'm pretty ready to give up on trying to let ProseMirror own the DOM structure.

tilgovi:
> Unless we want to pursue compiling a secondary ReactDOM renderer that we can flush synchronously, I'm pretty ready to give up on trying to let ProseMirror own the DOM structure.


And didn't you do some experiments with that from which you concluded even that had problems?

smoores-dev:
Alright, thanks @saranrapjs! If you think he'd be open to it, could be neat to have Marijn join the conversation here, too!

saranrapjs:
quoting Marijn:

> > e.g. where a bug fix requires a change to prosemirror-state and prosemirror-view, or prosemirror-tables and prosemirror-view (or something like that)
>
> This is very unlikely. I keep each of these packages backwards-compatible, by itself, and in cases where some old behavior has to be replaced I've set things up so that the old thing also remains available, but peer packages most to using the new thing.
>
> As for a full parallel implementation of prosemirror-view, I very much share the concern voiced in the PR that this will have to duplicate a lot of finnicky, fragile code (about 6000 lines), which will be a pain to get right and maintain.

I think without a plan to account for something like [`prosemirror-view`'s test suite](https://github.com/ProseMirror/prosemirror-view/tree/master/test), for example, I think I'd frame my POV as: we should only do this we can avoid breaking compatibility for library consumers that still want to use the mainline `prosemirror-view.EditorView`.

smoores-dev:
> we should only do this we can avoid breaking compatibility for library consumers that still want to use the mainline prosemirror-view.EditorView.

I don't totally understand what this would mean, I don't think. Are you suggesting two different APIs exposed by this library, the current one and the one proposed by this PR?

saranrapjs:
>  Are you suggesting two different APIs exposed by this library, the current one and the one proposed by this PR?

Is there any code on this PR that breaks `react-prosemirror`-managed NodeView functionality currently working with ProseMirror's `EditorView`? I see reimplementations of `EditorView`-isms, and a couple of classes (`NodeWrapper`, `TextNodeWrapper`) that maybe point to functionality that overlap with some of what `useNodeViews` does today — but is there code here that breaks the existing code used to manage NodeView placements? (maybe this would come further when you'd need to change the API "exposed" to NodeView React components to be more React-y — like some of what's in [Randall's snippet](https://github.com/nytimes/react-prosemirror/pull/47#discussion_r1272470869)?)

smoores-dev:
Nothing about the original API (the `<ProseMirror />` component and the `useNodViews` hook) is _broken_ by the new `<EditorView />` component, but it is _replaced_ by it. That is, aside from the `LayoutGroup` system, there's no overlap between the two interfaces, and I can't imagine how we could continue to expose both without it being very confusing for consumers (they're not _interchangeable_, you have to exclusively use one or the other). Essentially none of the old hooks, aside from maybe `useEditorState`, make much sense in the context of the `<EditorView />` component; and meanwhile there will need to be several new hooks (`useDomAtPos` and `useCoordsAtPos`, e.g.) that will not work/make sense in the old system.

Basically, I think we have to pick. I don't think this library can meaningfully support the actual EditorView class if we go with an all-React rendering approach. But I'm not at all opposed to attempting to run prosemirror-view's test suite (or at least a version of it) here to try to give us more confidence about compatibility.

I also want to be clear that I would much rather not be in a position where this feels like an approach we have to consider, but at this point we've been struggling to make React-based NodeViews interop with prosemirror-view for several years, and while I think react-prosemirror is the closest we've ever gotten, it still has some hurdles that feel increasingly gnarly. Basically, I'm saying that I don't think we're picking between "It works generally and also works with native EditorView" and "It probably works generally and doesn't work with native EditorView"; rather the first option is more like "It sort of works if you're very careful and we don't have a clear path toward it actually working robustly".

saranrapjs:
> Essentially none of the old hooks, aside from maybe useEditorState, make much sense in the context of the <EditorView /> component; and meanwhile there will need to be several new hooks (useDomAtPos and useCoordsAtPos, e.g.) that will not work/make sense in the old system

I think I'm not seeing how this is confusing — these seem like non-overlapping but mutually compatible code entrypoints we could maintain in parallel if we want to.

> Basically, I think we have to pick. 

I think I'm still not totally convinced that we do? To put it another way: if we want to make a bet that we can get a compatible implementation of `EditorView` working in React and stable enough that it provides value to users of this library such that they opt to switch to `react-prosemirror`'s `EditorView`, that seems like a totally worthwhile bet to make. But this library provides value now, even where as you note there are plenty of caveats centering around "it sort of works if you're very careful" that means that it's not where we might like it to be.

My feeling is that if there's a path to evaluating the stability of a `react-prosemirror` EditorView that doesn't force library consumers to choose between stability and robustness while we're still trying this out, we don't lose much in maintaining compatibility for some period of time. Unless you're picturing that this branch might be relatively long-lived while we work out all of the kinks? If you think it's reasonable that we wouldn't merge a branch like this until we'd have something like "compatibility with `prosemirror-view`'s test suite", that might be all I'm asking about here; I wouldn't feel as strongly about committing to avoiding breaking compatibility if we can commit to a way to measure our confidence.

smoores-dev:
Ah, I think I see where I was misunderstanding you. I had been interpreting your previous messages as a suggestion that we should _indefinitely_ maintain both interfaces.

> If you think it's reasonable that we wouldn't merge a branch like this until we'd have something like "compatibility with prosemirror-view's test suite", that might be all I'm asking about here; I wouldn't feel as strongly about committing to avoiding breaking compatibility if we can commit to a way to measure our confidence.

This is, in fact, exactly what my plan was, and I should have made that clearer upfront! I have no intention of merging this branch (or at least, exporting/documenting any of the new functionality) until we have a way to measure our confidence. I only opened this draft at this stage so that I could get a sense of whether this felt like a worthwhile endeavor to you all before I sink any more time into making it more robust!

tilgovi:
I don't want to rush this branch but I think this library provides very little value as it is now. It has some nice ideas, but without the stuff we've done internally at the Times to make simple node views work without wrappers, I'm skeptical that it's very useful to many people.

smoores-dev:
It sounds like there's enough interest in this approach to continue forward; I'm going to do my best to continue working on this direction, porting over code from prosemirror-view as needed (the coordsAtPos code for example is almost directly from prosemirror-view), and see if I can't match the prosemirror-view test suite as well, to add some confidence.

If anyone has any time to jam, my plan is to work on this every Friday for the foreseeable future, and I'd be very open to some pairing sessions!

smoores-dev:
@tilgovi (et al) the latest here is in sort of a nice place, I think, or at least heading there. I've directly copied over all of prosemirror-view's domcoords.ts, dom.ts, and browser.ts, and am simply passing the EditorView API implementation that we're building in EditorView.ts as the `view` argument to those implementations (suddenly very happy those are entirely implemented as standalone functions that take a view as an argument, rather than class methods!).

I've also copied over slightly truncated versions of the ViewDesc classes (removing anything related to `update` or `destroy`, since that's React's job now). I suspect that some parts of ViewDesc could be simplified if they relied on posToDesc, rather than the more elaborate position computation system that they're currently using, but it's working as is so I'm tempted to leave that for later, if at all.

There's a new `useView` hook that provides access to the view API. The restrictions are basically the same as the implementation on `main`: you can only safely access the view API in an effect or event handler (otherwise the parts of the view API that access the DOM may not be up-to-date).

I feel much, much, much better about this now that the ViewDesc classes and dom.ts/domcoords.ts code is working nearly as-is from prosemirror-view. I'd like to think about what `someProp` means in this system, and take a close look at prosemirror-view's `input.ts` and `selection.ts` and try to figure out how much of that needs to exist in this system, and how much should me managed with React instead.

smoores-dev:
@tilgovi @saranrapjs 

## Very big update

As I was going, my branch was increasingly looking like a chaotic fork of prosemirror-view, where I was hacking around, replacing the DOMObserver with a beforeinput listener, and replacing EditorView.update with my React components.

So I decided to lean into it, and spent some time today making it a _much less chaotic_ fork of prosemirror-view. The diff from prosemirror-view is now as small as I could get it:

1. I'm deferring the creation of the `docNodeView`, which prosemirror-view does in the `EditorView` constructor, to happen in some React useEffects.
2. I've overridden `EditorView.updateState`, `EditorView.updatePluginViews`, and `EditorView.setProps` so that they no longer try to mutate the DOM.
3. I've overridden the DOMObserver implementation so that it's only responsible for managing the selection, rather than actually observing the DOM with a MutationObserver

These first two changes allow React to be entirely responsible for translating the state into the DOM. The third change allows us to use a beforeInput observer for translating user inputs into ProseMirror transactions, rather than observing how the DOM changes and working backwards, which breaks React's rendering model (because the DOM is changed out from under it).

Similarly, I've limited the scope of prosemirror-view internals that the branch is using as much as possible. I'll have to do another pass to be sure, but I believe I'm now only using:

1. `selectionToDOM`, to ensure that react-prosemirror places the DOM selection in the same place that prosemirror-view would for any given ProseMirror Selection.
2. A very lightly forked implementation of `iterDeco`, to ensure that we render widget and node decorations in the same places that prosemirror-view does.
3. The ViewDesc class, and its subclasses. I have forked these for ease of use here, but I believe that could also use them as-is!

These changes are _so_ small that I think we should try to broach the conversation with Marijn about whether they could be upstreamed into prosemirror-view. With the exception of exposing iterDeco, selectionToDOM, and ViewDesc, the only changes that have to happen in prosemirror-view itself are:

1. Allow consumers of EditorView to set the docView themselves, somehow, rather than the default implementation
2. Allow consumers of EditorView to override the default DOMObserver
3. Allow consumers of EditorView to opt out of EditorView.updatePluginView (this is optional; we could work around this if we had to).

The other changes could be accomplished by subclassing EditorView to override method implementations, like updateState and setProps.

smoores-dev:
I'm going to try to summarize the state of this pull request; a lot has happened here since this was first suggested as an approach!

Firstly, the elephant in the room: This PR includes a fork of the `prosemirror-view` library, in the directory `prosemirror-view`. This is obviously not sustainable, and we shouldn't merge this PR with this fork! The diff between the fork and mainline `prosemirorr-view` is actually quite small, and I've called out everywhere that I actually changed the code of `prosemirror-view` with a `$$FORK` comment. This week I'm hoping to finish up work on going back to the `DOMObserver`, rather than the custom `beforeInput` handler currently on this branch (more on this below), and thereby make the diff even smaller. The explicit goal is to come up with a minimal tweak to the EditorProps interface that can be upstreamed into `prosemirror-view` that will support this work without an additional maintenance burden on this library _or_ `prosemirror-view`.

**TL;DR** - There's a fork of `prosemirror-view` here now, but we want to upstream the changes into the library before merging this branch. We won't merge this branch while it still has a fork of `prosemirror-view`; it's only there for development and experimentation.

## On to the actual approach:

This PR essentially replaces `prosemirror-view`'s `ViewDesc.updateChildren` with a React component tree. Everything else is the same, which is to say that this approach continues to use an actual instance of the `EditorView` for everything else that the `EditorView` does, including DOM observation, input detection and event handler execution, selection syncing, etc.

We do this by introducing some new React components, such as `DocNodeView`, `NodeView`, `TextNodeView`, `WidgetView`, and `MarkView`. Each of these is responsible for rendering some type of content represented by a view descriptor. That is, we have replaced the `updateChildren` method of each `ViewDesc` subclass with a React component with equivalent behavior.

There are some subtleties here, like the ProseMirror plugin that assigns unique keys to nodes to maintain React element stability across rerenders, that I'm happy to go into further detail about if anyone is interested.

## The current status

This PR has been brought to Marijn's attention [here](https://discuss.prosemirror.net/t/the-future-of-nytimes-react-prosemirror/5855). At his suggestion, I've been working on getting this approach working with `prosemirror-view`'s `DOMObserver` implementation, rather than the custom `beforeinput` handler approach that I originally tried here. There are, unfortunately, still widely used platforms (e.g. Android Chrome) that simply don't have well-behaved `beforeinput` behavior. Luckily this seems to be working, and I have a local diff that I will push later this week using `DOMObserver` that _mostly_ works, with a specific edge case around using composition events in a previously empty textblock.

My goal is to get this `DOMObserver` implementation up to 100% compliance with the test suite and current `prosemirror-view` behavior, and then propose a minimal change to the `prosemirror-view` library that will allow us to take this approach here!

src/prosemirror-view/viewdesc.ts
```diff
@@ -0,0 +1,1621 @@
...
+  }
+
+  get domAtom() { return false }
+
+  get ignoreForCoords() { return false }
+}
+
+// A widget desc represents a widget decoration, which is a DOM node
+// drawn between the document nodes.
+export class WidgetViewDesc extends ViewDesc {
```
saranrapjs, line 536:
```suggestion
// $$FORK: WidgetViewDesc not currently exported from prosemirror-view
export class WidgetViewDesc extends ViewDesc {
```

src/prosemirror-view/viewdesc.ts
```diff
@@ -0,0 +1,1621 @@
...
+    this.widget.type.destroy(this.dom)
+    super.destroy()
+  }
+
+  get domAtom() { return true }
+
+  get side() { return (this.widget.type as any).side as number }
+}
+
+export class CompositionViewDesc extends ViewDesc {
```
saranrapjs, line 572:
```suggestion
// $$FORK: CompositionViewDesc not currently exported from prosemirror-view
export class CompositionViewDesc extends ViewDesc {
```

src/prosemirror-view/viewdesc.ts
```diff
@@ -0,0 +1,1621 @@
...
+    return mut.type === 'characterData' && mut.target.nodeValue == mut.oldValue
+   }
+}
+
+// A mark desc represents a mark. May have multiple children,
+// depending on how the mark is split. Note that marks are drawn using
+// a fixed nesting order, for simplicity and predictability, so in
+// some cases they will be split more often than would appear
+// necessary.
+export class MarkViewDesc extends ViewDesc {
```
saranrapjs, line 598:
```suggestion
// $$FORK: MarkViewDesc not currently exported from prosemirror-view
export class MarkViewDesc extends ViewDesc {
```

src/prosemirror-view/viewdesc.ts
```diff
@@ -0,0 +1,1621 @@
...
+// and used by the view class.
+export function docViewDesc(doc: Node, outerDeco: readonly Decoration[], innerDeco: DecorationSource,
+                            dom: HTMLElement, view: EditorView): NodeViewDesc {
+  applyOuterDeco(dom, outerDeco, doc)
+  let docView = new NodeViewDesc(undefined, [], doc, outerDeco, innerDeco, dom, dom, dom)
+  if (docView.contentDOM) docView.updateChildren(view, 0)
+  return docView
+}
+
+export class TextViewDesc extends NodeViewDesc {
```
saranrapjs, line 895:
```suggestion
// $$FORK: TextViewDesc not currently exported from prosemirror-view
export class TextViewDesc extends NodeViewDesc {
```

src/prosemirror-view/viewdesc.ts
```diff
@@ -0,0 +1,1621 @@
...
+    if (this.dom != this.nodeDOM && (from == 0 || to == this.nodeDOM.nodeValue!.length))
+      this.dirty = NODE_DIRTY
+  }
+
+  get domAtom() { return false }
+}
+
+// A dummy desc used to tag trailing BR or IMG nodes created to work
+// around contentEditable terribleness.
+export class TrailingHackViewDesc extends ViewDesc {
```
saranrapjs, line 957:
```suggestion
// $$FORK: TrailingHackViewDesc not currently exported from prosemirror-view
export class TrailingHackViewDesc extends ViewDesc {
```

src/components/ChildNodeViews.tsx
```diff
@@ -0,0 +1,364 @@
...
+  innerPos: number;
+  childViews: Child[];
+};
+
+function InlineView({ innerPos, childViews }: SharedMarksProps) {
+  const { editorView } = useContext(EditorContext);
+  const editorState = useEditorState();
+  const reactKeys = useReactKeys();
+
+  const partitioned = childViews.reduce((acc, child) => {
```
saranrapjs, line 116: trying to wrap my head around this...this attempts to group like mark nodes and text widgets by mark type?

src/components/__tests__/ProseMirror.draw.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+    let sameAtEnd = 0;
+    while (
+      sameAtEnd < currentChildren.length &&
+      sameAtEnd < initialChildren.length &&
+      currentChildren[currentChildren.length - sameAtEnd - 1] ==
+        initialChildren[initialChildren.length - sameAtEnd - 1]
+    )
+      sameAtEnd++;
+    // $$FORK: Our node stability isn't _quite_ as robust
+    // as prosemirror-view's. The node adjacent to the one
```
saranrapjs, line 258: do you have a sense for why this is? like: is this about marks specifically, or generally about updates to DOM nodes adjacent to a given replacement?

smoores-dev: Ugh, unfortunately I don't remember the exact details of why this unit test is like one less than what it "should" be, but I did spend about an hour determining that it was fine haha. It's related to the way we do DOM stability/React key assignment generally. I'll try to explain, but take a look at the react keys plugin, too. DOM element stability is challenging in ProseMirror, because node content is an array, and the only way to preserve DOM stability in an array of elements in React is with unique keys. Buuuuttt... ProseMirror nodes don't have unique keys, nor any real unique identifiers. They can only really be uniquely identifiable by their position, which obviously can change as a result of any step.

So we have a plugin that stores a map from node position to a unique key for that node, and we do our best to keep those keys stable across transactions. This is actually quite good at node stability; it only runs into a small amount of trouble when adding new nodes, as this will inherently produce two positions that map backwards to the same previous position (making it challenging to determine which should actually receive the key from that position).

Anyway, something in there results in a teeny tiny amount of additional repaints in specific edge cases, but as the rest of this test suite demonstrates, this implementation gets really really close to prosemirror-view in terms of preserving DOM nodes.

src/components/ChildNodeViews.tsx
smoores-dev: Yeah this was gnarly and is probably not the most legible way to express it. There are some really complex rules that ProseMirror uses to flatten/simplify/group marks, and I wanted to be sure that we reproduced exactly the same behavior.

Here's what I wrote in the Gitter channel when I was working on this:

> A thing that I learned: Though inline model nodes are represented as flat nodes with marks, when possible, adjacent nodes with the same outer decorations and marks are actually merged in the HTML output. So for example:
>
>```
>doc
>  p
>    "bold and italic" #text [em, strong]
>    "italic" #text [em]
>```
>
> Is rendered as:
>
>```html
><div>
>  <p>
>    <em>
>      <strong>bold and italic</strong>
>      italic
>    </em>
>  </p>
></div>
>```

smoores-dev: Ok well they're not THAT complex, they're just non-trivial to express with React haha

src/prosemirror-view/viewdesc.ts
smoores-dev: Good call. I actually need to take a minute at some point and confirm that either:

1) ProseMirror just works fine if you use custom ViewDesc classes that match the interface it cares about

OR

2) The existing ViewDesc classes will work for what we need.

I think that _both_ of these are likely to be true, so which one we go with will be more determined by how comfortable Marijn ends up feeling about exporting these classes. But again, I haven't verified yet.

src/components/__tests__/ProseMirror.draw.test.tsx
saranrapjs: Got it. So it's maybe not specific marks, but instead general to React elements at the end of a list, or something like that? As you said doesn't seem super critical, though it might be interesting at some point to poke into when/why it does/doesn't trigger, if only to understand if there's some way we could improve the React element keying to reduce the odds of repaint

smoores-dev: I feel confident that it's not specific to marks, and yeah more like something to do with something like what you said. And I think documenting that is absolutely a good call, as well as seeing if it can be improved even further!

tilgovi: I also wouldn't be surprised if there's places where stability is better than in the standard ProseMirror editor view. Although, stability _in theory_ isn't necessarily the same as stability in common practice. There might be cases this handles better but that don't really occur in practice. I remember thinking the way ProseMirror handles this is pretty clever, something like matching prefix and suffixes of the sibling lists, if I remember correctly, no?

smoores-dev:
Heads up: I just quite dramatically rebased the git history of this branch, which was at 77 commits and had gotten completely unwieldy. For folks taking a look at this branch for the first time: I recommend looking through the commits in order to get a sense of what's going on!

plaisted:
Is there a `getting started` guide or similar doc for this branch? Is this published anywhere as a package yet?

smoores-dev:
@plaisted yeah! The README on this branch has been updated to reflect the changes to the APIs that this branch makes, and it's published to npm under the `next` tag (the latest version is `0.7.0-next.5`).

src/components/TextNodeView.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export class TextNodeView extends Component<Props> {
+  private viewDescRef: null | TextViewDesc = null;
+
+  updateEffect() {
+    const { decorations, siblingDescriptors, node } = this.props;
+    // There simply is no other way to ref a text node
+    // eslint-disable-next-line react/no-find-dom-node
+    const dom = findDOMNode(this);
```
sethfair, line 25: I am getting the following warning in the console.  Im thinking this is due to how it interacts with ProseMirror under the hood?

`Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of TextNodeView which is inside StrictMode. Instead, add a ref directly to the element you want to reference.`

smoores-dev: Hi @sethfair! Until React introduces a mechanism for obtaining a ref of a text node, we're stuck using a deprecated function, `findDOMNode`. This will not introduce any issues, but it does unfortunately result in this warning being printed; you can read more about the deprecation here: https://react.dev/reference/react-dom/findDOMNode#alternatives. Note that the authors don't seem to plan on removing `findDOMNode` until they've provided alternatives for all of its use cases, so I feel comfortable using it for this scenario, which has no alternative.

src/hooks/useEditorState.ts
```diff
@@ -7,7 +7,7 @@ import { EditorContext } from "../contexts/EditorContext.js";
  * Provides access to the current EditorState value.
  */
 export function useEditorState(): EditorState | null {
-  const { editorState } = useContext(EditorContext);
+  const { state: editorState } = useContext(EditorContext);
```
saranrapjs, line 10: Would you mind if I pulled the change in https://github.com/nytimes/react-prosemirror/pull/84 here too? I'm realizing we've relied on non-nullability in a bunch of places now, and want to remove that from the diffs in getting an understanding of the impact of this branch on some of our code.

smoores-dev: Turns out this was already non-nullable everywhere, this hook was just incorrectly typed! I just pushed the update

tilgovi (approved):

PR #49, opened 2023-09-11
Modify useEditorView to flushSync all dispatched transactions

ilyaGurevich:
This change ensures a flushSync-only behavior when dispatching transactions. [Because react 18 automatically batches state updates](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching), we can also remove references in the code where we used unstable_batchedUpdates in general.

### DIFF
--- a/.yarn/versions/290b480a.yml
+++ b/.yarn/versions/290b480a.yml
@@ -1,0 +1,2 @@
+releases:
+  "@nytimes/react-prosemirror": patch
--- a/.yarn/versions/290b480a.yml
+++ b/.yarn/versions/290b480a.yml
@@ -1,2 +1,2 @@
 releases:
-  "@nytimes/react-prosemirror": patch
+  "@nytimes/react-prosemirror": minor
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,3 +1,5 @@
+nodeLinker: pnp
+
 changesetIgnorePatterns:
   - "**/*.{test}.{js,ts,jsx,tsx}"
   - "!(.yarn/versions/**|src/**/*|package.json|.swcrc*)"
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,5 +1,3 @@
-nodeLinker: pnp
-
 changesetIgnorePatterns:
   - "**/*.{test}.{js,ts,jsx,tsx}"
   - "!(.yarn/versions/**|src/**/*|package.json|.swcrc*)"
--- a/src/hooks/useEditorView.ts
+++ b/src/hooks/useEditorView.ts
@@ -1,18 +1,23 @@
 import type { EditorState, Transaction } from "prosemirror-state";
 import { EditorView } from "prosemirror-view";
 import type { DirectEditorProps } from "prosemirror-view";
-import { useLayoutEffect, useState } from "react";
+import { useLayoutEffect, useRef, useState } from "react";
 import { unstable_batchedUpdates as batch } from "react-dom";
 
 import { useForceUpdate } from "./useForceUpdate.js";
 
-function withBatchedUpdates<This, T extends unknown[]>(
-  fn: (this: This, ...args: T) => void
+function withConditionalFlushUpdates<This, T extends unknown[]>(
+  fn: (this: This, ...args: T) => void,
+  view: EditorView | null
 ): (...args: T) => void {
   return function (this: This, ...args: T) {
-    batch(() => {
+    if (view?.composing) {
+      batch(() => {
+        fn.call(this, ...args);
+      });
+    } else {
       fn.call(this, ...args);
-    });
+    }
   };
 }
 
@@ -42,9 +47,10 @@ export type EditorProps = Omit<DirectEditorProps, "state"> & EditorStateProps;
  * the Editor View unmodified after we upgrade to React 18, which batches every
  * update by default.
  */
-function withBatchedDispatch(
+function withConditionalFlushDispatch(
   props: EditorProps,
-  forceUpdate: () => void
+  forceUpdate: () => void,
+  view: EditorView | null
 ): EditorProps & {
   dispatchTransaction: EditorView["dispatch"];
 } {
@@ -55,10 +61,11 @@ function withBatchedDispatch(
         this: EditorView,
         tr: Transaction
       ) {
-        const batchedDispatchTransaction = withBatchedUpdates(
-          props.dispatchTransaction ?? defaultDispatchTransaction
+        const conditionallyFlushedDispatch = withConditionalFlushUpdates(
+          props.dispatchTransaction ?? defaultDispatchTransaction,
+          view
         );
-        batchedDispatchTransaction.call(this, tr);
+        conditionallyFlushedDispatch.call(this, tr);
         forceUpdate();
       },
     },
@@ -79,20 +86,22 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
   props: EditorProps
 ): EditorView | null {
   const [view, setView] = useState<EditorView | null>(null);
+  const editorPropsRef = useRef(props);
 
   const forceUpdate = useForceUpdate();
 
-  const editorProps = withBatchedDispatch(props, forceUpdate);
-
-  const stateProp = "state" in editorProps ? editorProps.state : undefined;
+  const stateProp =
+    "state" in editorPropsRef.current
+      ? editorPropsRef.current.state
+      : undefined;
 
   const state =
-    "defaultState" in editorProps
-      ? editorProps.defaultState
-      : editorProps.state;
+    "defaultState" in editorPropsRef.current
+      ? editorPropsRef.current.defaultState
+      : editorPropsRef.current.state;
 
   const nonStateProps = Object.fromEntries(
-    Object.entries(editorProps).filter(
+    Object.entries(editorPropsRef.current).filter(
       ([propName]) => propName !== "state" && propName !== "defaultState"
     )
   );
@@ -120,14 +129,14 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
         new EditorView(
           { mount },
           {
-            ...editorProps,
+            ...editorPropsRef.current,
             state,
           }
         )
       );
       return;
     }
-  }, [editorProps, mount, state, view]);
+  }, [props, mount, state, view]);
 
   useLayoutEffect(() => {
     view?.setProps(nonStateProps);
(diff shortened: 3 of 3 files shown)


src/hooks/useEditorView.ts
```diff
@@ -137,5 +146,13 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
     if (stateProp) view?.setProps({ state: stateProp });
   }, [view, stateProp]);
 
+  useLayoutEffect(() => {
+    editorPropsRef.current = withConditionalFlushDispatch(
+      props,
+      forceUpdate,
+      view
+    );
+  }, [props, view]);
```
smoores-dev, line 155: Looks like we're missing forceUpdate here. It has a stable identity, so it's safe to add to the dependencies

ilyaGurevich: aah, needed to restart my eslint server on vscode, pushing up

smoores-dev: Does this layout effect need to be declared before the one where we update the view? As it is, I believe the view will be updated with the _old_ props on a props change, because the ref won't have been updated yet when that effect runs

ilyaGurevich: That sounds right to me 👍

smoores-dev:
@ilyaGurevich I'm not sure the ref is getting us what we want here! Is it possible that we just want to useMemo the call to `withBatchedUpdates`? It seems like the new layout effect hook is updating the ref any time the dependencies change, which should be the same as how useMemo works.

smoores-dev:
Also, how would you feel about trying to add a unit test to capture this use case? Feels edge-case-y enough to earn one!

tilgovi:
I'm confused and want to investigate a little bit more. Using `flushSync` actually creates a batch boundary. We do it so that if `dispatchTransaction` triggers a React state change that React doesn't flush any effects of that state change until we unwind the stack. We introduced this to prevent nested Redux dispatches when components have effects that dispatch transactions (like something dispatching on mount).

So, using `flushSync` prevents this dispatch from getting rolled up into a larger batch, to flush all together, but it doesn't avoid batching. I'd be shocked if the IME composition case is part of a larger batch that isn't otherwise flushed synchronously. At least in React 17, event handling always happened within a batch and that batch flushed synchronously.

tilgovi:
I found one bit of interesting info here. The actual dispatch in the case of typing is coming from ProseMirror reading a DOM mutation, not from an event. On React 17, that would be unbatched. On React 18, that'd be batched but not synchronous.

I have suggested previously that we should always flush sync, since editor input is high priority for proper user experience. The only reason I hesitated was that I was afraid we had code at NYT that had components dispatching on mount, which is not a legal place to do a `flushSync`.

I'd be happy to synchronously flush unconditionally here, and work out issues with dispatching at bad lifecycle moments downstream.

tilgovi:
In other words, because ProseMirror uses `MutationObserver` rather than input events, letting the browser handle most editing operations, resulting React state changes have always been unbatched, which means they've always applied synchronously. We added the batching to ensure that we didn't cause nested dispatches when components dispatch in lifecycle hooks. Going a step further, we could flush that batch synchronously, which would be ideal for an editor.

I think this is on the right track, but doesn't go far enough, and the justification isn't quite correct. We don't want to opt out of batching. We want to be sure the batch commits synchronously.

src/hooks/useEditorView.ts
```diff
@@ -1,18 +1,37 @@
+import { useForceUpdate } from "./useForceUpdate.js";
```
ilyaGurevich, line 1: @smoores-dev any idea here why github actions check wouldn't be aligned with my local eslint/vscode lint configuration?  I'm not seeing this error on my end. I'm trying to poke around with whether or not I'm using some global eslint configuration instead of the local one, but I can't see why that would be the case.

ilyaGurevich:
> I'd be happy to synchronously flush unconditionally here, and work out issues with dispatching at bad lifecycle moments downstream.

I'm ok doing this as well, I think it might also make this overall logic simpler to reason around without needing to specifically understand what composition is in general. What sorts of examples of bad lifecycles moments would you expect in this case? Like a bad case of `useEditorEffect`?

src/hooks/useEditorView.ts
smoores-dev: Hmmmm. There shouldn't be a `node_modules/eslint` in this project at all; it's using plug n play, which doesn't install packages in node modules. That's why there's a `.vscode/settings.json` that configures `"eslint.nodePath": ".yarn/sdks"`; so that the VS Code ESLint extension will use the PnP loader to load the ESLint package.

smoores-dev: To run the checks locally, you should be able to run `yarn check`

smoores-dev: But it should also be working with VS Code; it does for me!

ilyaGurevich: hmm, I ran `yarn` and I ended up going a node_nodules folder in my directory, looks like I somehow got opted out of PnP for some reason

smoores-dev: Does the `node_modules` directory actually have packages in it, or just caches like `.cache` and `.vite`? The only way I could imagine yarn not using PnP is if you're using a very old version of yarn. What do you see when you run `yarn --version`?

ilyaGurevich: I re-installed everything and `yarn check` gave me a whole host of import order issues for some reason 
```
<img width="1298" alt="image" src="https://github.com/nytimes/react-prosemirror/assets/19535560/4c07f871-09a3-47b8-ae31-437cae193e7a">
```

ilyaGurevich: > . What do you see when you run `yarn --version`?

yarn --version gives me `3.4.1`

ilyaGurevich: image

smoores-dev: Yeah, I'm not sure what's going on here. Has your yarn.lock changed somehow? Did you install newer versions of eslint or the import plugin?

ilyaGurevich: I even ran vscode in a devcontainer and eslint gives me the same result.

ilyaGurevich: So I just re-cloned the repo. Here's the steps I took (I didn't even open vscode)
- nvm use v18
- yarn 
- yarn check

I get a huge output of import errors, FWIW I don't get the same issue in a .devcontainer version, I just am not sure why they would be different

tilgovi:
> Like a bad case of useEditorEffect?

Yeah. It's risky to dispatch transactions in `useEditorEffect` because it can lean to infinite dispatch loops. Using correct dependencies or conditionals can avoid that. I was afraid we would have cases where it is correct, though, which is why I hesitated to use `flushSync()` originally. Batching updates was the gentler version of `flushSync()`.

> I think I'm also still somewhat confused on why:
>
> * we were purposefully batching dispatch transactions in react 17
> * flushSync'ing these transactions is acceptable now in react 18

We were purposefully batching transactions so that when a dispatch causes a state change (generally via Redux), it doesn't cause React to invoke any component lifecycle hooks until after we return from the call to `dispatchTransaction`. That way we avoid nested dispatches.

Using `flushSync()` would have been acceptable on React 17, too, I just wasn't sure it was acceptable from a developer ergonomics perspective because it'll throw warnings in the console if you dispatch from a component lifecycle update.

tilgovi:
Just to reiterate/clarify: `flushSync() == `commit(batch())`. It is stronger.

.yarnrc.yml
```diff
@@ -1,3 +1,5 @@
+nodeLinker: pnp
```
tilgovi, line 1: This repository does not use Plug'n'Play.

.yarn/versions/290b480a.yml
```diff
@@ -0,0 +1,2 @@
+releases:
+  "@nytimes/react-prosemirror": patch
```
tilgovi, line 2: I'd just call this minor. I'd even say it would be major, but we're pre-1.0.

.yarnrc.yml
ilyaGurevich: I believe it is based on this [thread](https://github.com/nytimes/react-prosemirror/pull/49#discussion_r1321867957), in some cases though you'll need to explicitly define it.

tilgovi: Oh, my confusion! It's using Plug'N'Play but not zero installs.

ilyaGurevich:
> Yeah. It's risky to dispatch transactions in `useEditorEffect` because it can lean to infinite dispatch loops

I'd be curious about making a custom eslint configuration here to provide warnings when you do stuff like this, similar to calling setState in a useEffect which is discouraged sometimes

.yarnrc.yml
tilgovi: This is actually the default, so you shouldn't need it.

tilgovi:
> I'd be curious about making a custom eslint configuration here to provide warnings when you do stuff like this, similar to calling setState in a useEffect which is discouraged sometimes

Yeah, it'd be nice, but I think it's low on the priorities for this repo. Right now, we're not even sure we have the implementation at all how we'd like it.

.yarnrc.yml
ilyaGurevich: I had similar issues ensuring plug'n'play was enabled on my machine when working on the Oak repo, I had to explicitly define it in this repo for the same reason I had to define it in the other one. My version of yarn auto-defaults to non-plug-n-play (for one reason or another).

tilgovi: Do you have a global yarn rc somewhere that specifies nodeModules?

tilgovi: `yarn config --why` might help.

ilyaGurevich: My `~/.yarnrc` is blank unfortunately, so not sure where that default would be coming from.

ilyaGurevich: > yarn config --why
huh, I had a floating .yarnrc.yml file in a random folder for some reason that was the default, i removed it so should be fine

src/hooks/useEditorView.ts
smoores-dev: Oh... interesting! Do you have a `~/.yarnrc.yml` that has `nodeLinker: node-modules`? That's the only reason I could imagine it not defaulting to pnp

smoores-dev: (I see the other thread now, glad we figured it out!)

smoores-dev:
Thanks for this, @ilyaGurevich! Would you mind rebasing on top of the latest on `main` so that the unit tests get run on this branch?

tilgovi (approved):

smoores-dev:
@ilyaGurevich do you need a new npm release for this?

ilyaGurevich:
> @ilyaGurevich do you need a new npm release for this?

Not yet, I have a follow-up PR I'd like to open and discuss first before doing npm release!

PR #51, opened 2023-09-11
fix: ordering of layout effect not running properly on compose events

ilyaGurevich:
After switching to flush-sync, compose events were still not working immediately, this is due to the ordering of the useLayoutEffect for the state prop. Prop state should be set first before non-state props.


ilyaGurevich:
Closed as I'm realizing the other PR is still opened

PR #52, opened 2023-09-11
Rearrange props setting in useEditorView

ilyaGurevich:
Composition events were still encountering issues (even with the flushSync changes). An initial fix was identified which was to change the order of setting the state and nonState props in their respective layout effects, but I'm not entirely sure they should be separated. We can combine them to improve simplicity. SetProps should already be efficient enough as an improvements over `update`, we may be calling it more frequently but it feels relatively low cost.

### DIFF
--- a/src/components/__tests__/ProseMirror.test.tsx
+++ b/src/components/__tests__/ProseMirror.test.tsx
@@ -12,6 +12,7 @@ import {
   teardownProseMirrorView,
 } from "../../testing/setupProseMirrorView.js";
 import { ProseMirror } from "../ProseMirror.js";
+import { simulateType } from "../utils/textKeyboardSimulator.js";
 
 describe("ProseMirror", () => {
   beforeAll(() => {
@@ -139,6 +140,63 @@ describe("ProseMirror", () => {
     expect(screen.getAllByTestId("paragraph").length).toBeGreaterThanOrEqual(1);
   });
 
+  it("supports composition events", async () => {
+    const schema = new Schema({
+      nodes: {
+        text: {},
+        paragraph: { content: "text*" },
+        doc: { content: "paragraph+" },
+      },
+    });
+    const editorState = EditorState.create({ schema });
+
+    function Paragraph({ children }: NodeViewComponentProps) {
+      return <p data-testid="paragraph">{children}</p>;
+    }
+
+    const reactNodeViews = {
+      paragraph: () => ({
+        component: Paragraph,
+        dom: document.createElement("div"),
+        contentDOM: document.createElement("span"),
+      }),
+    };
+
+    function TestEditor() {
+      const { nodeViews, renderNodeViews } = useNodeViews(reactNodeViews);
+      const [mount, setMount] = useState<HTMLDivElement | null>(null);
+
+      return (
+        <ProseMirror
+          mount={mount}
+          state={editorState}
+          dispatchTransaction={function (this: EditorView, tr) {
+            // We have to wrap the update in `act` to handle all of
+            // the async portal registering and component rendering that
+            // happens "out of band" because it's triggered by ProseMirror,
+            // not React.
+            act(() => this.updateState(this.state.apply(tr)));
+          }}
+          nodeViews={nodeViews}
+        >
+          <div data-testid="editor" ref={setMount} />
+          {renderNodeViews()}
+        </ProseMirror>
+      );
+    }
+
+    render(<TestEditor />);
+
+    const editor = screen.getByTestId("editor");
+
+    await editor.focus();
+    simulateType("ó");
+
+    expect(editor.textContent).toBe("ó");
+    // Ensure that ProseMirror really rendered our Paragraph
+    // component, not just any old <p> tag
+    expect(screen.getAllByTestId("paragraph").length).toBeGreaterThanOrEqual(1);
+  });
   afterAll(() => {
     teardownProseMirrorView();
   });
--- a/src/hooks/useEditorView.ts
+++ b/src/hooks/useEditorView.ts
@@ -71,10 +71,18 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
 
   const editorProps = withFlushedDispatch(props, forceUpdate);
 
+  const stateProp = "state" in editorProps ? editorProps.state : undefined;
+
   const state =
     "defaultState" in editorProps
       ? editorProps.defaultState
       : editorProps.state;
+
+  const nonStateProps = Object.fromEntries(
+    Object.entries(editorProps).filter(
+      ([propName]) => propName !== "state" && propName !== "defaultState"
+    )
+  );
 
   useLayoutEffect(() => {
     return () => {
@@ -109,8 +117,12 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
   }, [editorProps, mount, state, view]);
 
   useLayoutEffect(() => {
-    view?.setProps(editorProps);
-  }, [view, editorProps]);
+    if (stateProp) view?.setProps({ state: stateProp });
+  }, [view, stateProp]);
+
+  useLayoutEffect(() => {
+    view?.setProps(nonStateProps);
+  }, [view, nonStateProps]);
 
   return view;
 }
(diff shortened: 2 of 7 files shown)


src/hooks/useEditorView.ts
```diff
@@ -117,12 +109,8 @@ export function useEditorView<T extends HTMLElement = HTMLElement>(
...
 
   useLayoutEffect(() => {
-    view?.setProps(nonStateProps);
-  }, [view, nonStateProps]);
-
-  useLayoutEffect(() => {
-    if (stateProp) view?.setProps({ state: stateProp });
-  }, [view, stateProp]);
+    view?.setProps(editorProps);
+  }, [view, editorProps]);
```
smoores-dev, line 113: I think this might cause subtle issues. If I remember correctly, ProseMirror will execute side effects when various props are set (e.g. destroying and recreating all node views when the nodeViews prop is set), even if the reference is the same. We separated out the state update because the state prop updates very often as a user edits the document, and we don't want to trigger these side effects every time the state is updated.

smoores-dev: Actually, I think that the failing test might even be an example of what I'm talking about here (it might be something else, though)

smoores-dev:
Also, @ilyaGurevich, I just wanted to make sure you all were aware of #47, which is going to pretty dramatically shake up the underlying implementation for all of this!

src/hooks/useEditorView.ts
tilgovi: I thought we worked with Marijn to fix some of those issues upstream, so that even if we set the props things should be efficient as long as the value of the individual props didn't change.

ilyaGurevich: If they remain separated, is the initial approach of setting the state props before the nonState props in the layout effect alright in this case?

ilyaGurevich:
> Also, @ilyaGurevich, I just wanted to make sure you all were aware of #47, which is going to pretty dramatically shake up the underlying implementation for all of this!

I saw! Can't wait for this, but we'd like to get a version of this fix out for now (to also mimic in Oak)

src/hooks/useEditorView.ts
tilgovi: I think doing it all at once is the safest, so if we do need to be efficient we can do it like we have in production code before the upstreamed changes to prosemirror-view that fixed inefficiencies:

```
    const { state: oldState, ...oldProps } = view.props;
    const { state: newState, ...newProps } = props;
    if (!isShallowEqual(oldProps, newProps)) {
      view.update(props);
    } else {
      view.updateState(props.state);
    }
```

smoores-dev:
@ilyaGurevich would you be down to try to write a unit test for this, so that we can feel more confident about this fix?

src/hooks/useEditorView.ts
smoores-dev: Isn't the layout effect setup now essentially just doing that shallow equal comparison for us?

smoores-dev: ... ha. I just realized that we're actually calling view.setProps on literally every render, right now, because nonStateProps isn't memoized. So actually I'm not sure @ilyaGurevich's initial change would have caused issues.

Also, I'm not sure why we use `setProps` here rather than `update`/`updateState`? @tilgovi do you remember?

tilgovi: `update` and `updateState` allow partial updates, I think, but we have the whole props.

ilyaGurevich: Hmm ok, what's the preferred solution here? It looks like we have multiple approaches all relatively doing similar things:
1. rearrange order (with memoized nonStateProps)
```
  useLayoutEffect(() => {
    if (stateProp) view?.setProps({ state: stateProp });
  }, [view, stateProp]);
  
  useLayoutEffect(() => {
    view?.setProps(nonStateProps);
  }, [view, nonStateProps]);
  ```
  
2. Unify props setting
```
  useLayoutEffect(() => {
    view?.setProps(editorProps);
  }, [view, editorProps]);
```

3. Use the older version in production
```
    const { state: oldState, ...oldProps } = view.props;
    const { state: newState, ...newProps } = props;
    if (!isShallowEqual(oldProps, newProps)) {
      view.update(props);
    } else {
      view.updateState(props.state);
    }
```

**Comment by ilyaGurevich** [comment]
> would you be down to try to write a unit test for this, so that we can feel more confident about this fix?

If we go with this approach: 
```
const nonStateProps = useMemo(
  () =>
    Object.fromEntries(
      Object.entries(editorProps).filter(
        ([propName]) => propName !== "state" && propName !== "defaultState"
      )
    ),
  [editorProps]
);
 
<  ..............>

useLayoutEffect(() => {
  if (stateProp) view?.setProps({ state: stateProp });
}, [view, stateProp]);

useLayoutEffect(() => {
  view?.setProps(nonStateProps);
}, [view, nonStateProps]);
```
Besides the existing "Hello World" test how best do you think we can test this? I tried recreating compositionStart/Update/End tests but it seems unlikely that we'll get that functionality available via user-events.

**Comment by smoores-dev** [comment]
Yeah I hear you. Someone [here](https://github.com/testing-library/user-event/issues/1097) has documented the events that get fired when using a composition to type á on an Icelandic keyboard; could we try just firing exactly these events in this order, and verify that we get the correct character at the end? The complaint in the linked issue doesn't seem to be that it wouldn't work, only that it's quite verbose.

**Comment by smoores-dev** [comment]
Also don't forget that you can run the demo locally with `yarn demo` to verify your changes manually. I would definitely love to have an automated test for this, because it feels very easy to accidentally break, but if we absolutely can't figure out how to test it, let's definitely verify manually!

**Comment by ilyaGurevich** [comment]
> Yeah I hear you. Someone [here](https://github.com/testing-library/user-event/issues/1097) has documented the events that get fired when using a composition to type á on an Icelandic keyboard; could we try just firing exactly these events in this order, and verify that we get the correct character at the end? The complaint in the linked issue doesn't seem to be that it wouldn't work, only that it's quite verbose.

I ended up reaching out to the commenter who very graciously sent me his code block for this bit of work. I pushed up a ts-if-ied version of this code and added a test for it

**Comment by smoores-dev** [comment]
Oh wow, that's amazing. Good thought to ask! And just to triple check; the test you added fails without this change, right?

**Comment by ilyaGurevich** [comment]
> Oh wow, that's amazing. Good thought to ask! And just to triple check; the test you added fails without this change, right?

Hmm, not it doesn't look like it fails in this repo, although without the switch the behavior is not working in oak-side

**Comment by smoores-dev** [comment]
Have you been able to reproduce the issue that you're seeing in Oak in the demo editor in this repo?

**Comment by ilyaGurevich** [comment]
> 

uuf, the issue is currently happening in the demo and with the change it is still occurring 😞

File: src/hooks/useEditorView.ts (hunk 1 above)
Comment (tilgovi) [reply src/hooks/useEditorView.ts]: I think unified props setting is the most sensible, and I'd be curious about the test breakage.

File: src/components/__tests__/ProseMirror.test.tsx [hunk 2]
```diff
@@ -139,6 +140,63 @@ describe("ProseMirror", () => {
...
+      return (
+        <ProseMirror
+          mount={mount}
+          state={editorState}
+          dispatchTransaction={function (this: EditorView, tr) {
+            // We have to wrap the update in `act` to handle all of
+            // the async portal registering and component rendering that
+            // happens "out of band" because it's triggered by ProseMirror,
+            // not React.
+            act(() => this.updateState(this.state.apply(tr)));
```
tilgovi, line 178: I would think we would want to wrap the `simulateType` call in act, instead, no?

smoores-dev: I don't think so; simulateType doesn't synchronously trigger the React state update. It changes the DOM, which ProseMirror observes, and then runs dispatchTransaction, which finally updates the state

ilyaGurevich:
@smoores-dev you mentioned there being dramatic changes to this file in the future anyway, is it worth keeping this PR open at its current time? It is more useful to wait until the overall editorview changes are integrated?

smoores-dev:
If Oak isn't currently using this package (and has a fix that's working for them), and we don't have a fix that we're confident in, I think we shouldn't worry about it. There are known issues with the way this package currently interacts with ProseMirror's DOMObserver that we hope to resolve with #47.

ilyaGurevich:
@smoores-dev that sounds good to me, I'll go ahead and close this PR

PR #62, opened 2023-11-14
fixes errors thrown when React unmounts the contentDOMWrapper

saranrapjs:
This PR proposes a fix to #42 within the current architecture (where we create wrapping nodes and React portals for node views).

The inspiration for this approach comes from a previous implementation where we had to do something like what happens on this branch: at the moment that a NodeView's `destroy` hook is being called, we re-attach the parts of the DOM which React controls in roughly the locations where React expects them for its cleanup phase.

### DIFF
--- a/src/nodeViews/createReactNodeViewConstructor.tsx
+++ b/src/nodeViews/createReactNodeViewConstructor.tsx
@@ -174,7 +174,9 @@ export function createReactNodeViewConstructor(
                   // parent so that later we can reassemble the DOM hierarchy
                   // React expects when cleaning up the ContentDOMWrapper element
                   if (nextContentDOMWrapper?.parentNode) {
-                    setContentDOMParent(nextContentDOMWrapper.parentNode as HTMLElement);                    
+                    setContentDOMParent(
+                      nextContentDOMWrapper.parentNode as HTMLElement
+                    );
                   }
                 }}
               />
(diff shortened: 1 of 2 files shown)


src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -252,6 +264,14 @@ export function createReactNodeViewConstructor(
         return false;
       },
       destroy() {
+        // React expects the contentDOMParent to be a child of the
+        // DOM element where the portal was mounted, but in some situations
+        // contenteditable may have already detached the contentDOMParent
+        // from the DOM. Here we attempt to reassemble the DOM that React
+        // expects when cleaning up the portal.
+        if (componentRef?.contentDOMParent) {
+          this.dom.appendChild(componentRef?.contentDOMParent);
```
tilgovi, line 273: nit: the `?.` here shouldn't be necessary.

tilgovi (approved):
This seems reasonable to me. Hacky, but that's the territory.

smoores-dev:
@saranrapjs you can resolve the version check issue with `yarn version patch --deferred`. It should make a new yaml file in the `.yarn/versions` directory. And I agree with @tilgovi, this looks good!

smoores-dev (approved):

PR #64, opened 2023-11-14
0.4.0

saranrapjs:
This tags a new version reflecting some of the latest commits on `main`. I'm...not 100% sure how to make sure that, when this merges, we get a new version in `npm`. It looks like we don't auto-publish to `npm` (which is fine) but I'll check in w/ other folks to try to get the process documented, if it's not already.

### DIFF
--- a/.yarn/versions/5245b9ba.yml
+++ b/.yarn/versions/5245b9ba.yml
@@ -1,2 +1,0 @@
-releases:
-  "@nytimes/react-prosemirror": patch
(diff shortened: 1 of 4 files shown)


smoores-dev:
I've been publishing manually after releases merge (happy to do that here if you'd like!)

If you want to get permission to publish yourself, there are docs on how to add yourself as a publisher for the `@nytimes` npm org on the internal OSS docs site.

saranrapjs:
> I've been publishing manually after releases merge (happy to do that here if you'd like!)

If you're game to do this for this merge, I can get setup to do this in future, in parallel!

smoores-dev (approved):

smoores-dev:
All set! https://www.npmjs.com/package/@nytimes/react-prosemirror/v/0.4.0

PR #78, opened 2024-01-22
feat(editor): add before and after slots to editor component

konstantinmuenster:
While migrating our codebase to the newer ReactEditorView, I realized that there is no convenient way to add a editor toolbar above the rendered document. The children of the `ProseMirror` component get inserted after the actual editor view.

A simple fix could be to define slots, before and after the document. This might even be a clearer wording than `children` since nested components aren't inserted as actual children.

But open for other solutions / ideas :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,8 +381,8 @@ function ProseMirrorEditor() {
 ```tsx
 type ProseMirror = (
   props: {
-    mount: HTMLElement;
-    children: ReactNode;
+    slotBefore: ReactNode;
+    slotAfter: ReactNode;
   } & DirectEditorProps &
     ({ defaultState: EditorState } | { state: EditorState })
 ) => JSX.Element;
--- a/src/components/ProseMirror.tsx
+++ b/src/components/ProseMirror.tsx
@@ -67,15 +67,17 @@ export type EditorProps = Omit<
 
 export type Props = EditorProps & {
   className?: string;
-  children?: ReactNode;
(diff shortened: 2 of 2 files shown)


smoores-dev:
Thanks @konstantinmuenster, this makes a lot of sense! Another approach I'm considering is something like:

```tsx
import { ProseMirrorDoc, ProseMirror } from '@nytimes/react-prosemirror';

...

      <ProseMirror
        state={state}
        dispatchTransaction={function (tr) {
          setState((prev) => prev.apply(tr));
        }}
      >
        {/** any children before */}
        <ProseMirrorDoc />
        {/** any children after */}
      </ProseMirror>
```

Which would give back the full flexibility that the previous implementation had, allowing you to not just add children before and after, but nest the document element as needed, etc. This is probably where the `as` prop would move to, as well (currently on the `ProseMirror` component).

What do you think, any preference?

konstantinmuenster:
That's even better. How would you apply necessary props to the `ProseMirrorDoc` component? via context?

If you had it on your list anyway, feel free to discard my changes and add your approach instead 👍

smoores-dev:
Yup, exactly! We'll add a context that provides the necessary props for `DocNodeView`, and then `ProseMirrorDoc` will consume that context and render the `DocNodeView`

smoores-dev:
Sorry for the delay, just published! @konstantinmuenster

PR #81, opened 2024-01-25
Menu demo

coolov:
Hello!

This adds a menu to the demo. The intention is to show how to synchronize state from ProseMirror to React outside of the `<ProseMirror>` component hierarchy. Am I doing this right?

Features:
- Toggle bold and italic using React buttons that dispatches Transactions to the React managed EditorState.
- Derive the "active state" of the buttons using cursor positions that are derived from the EditorState

https://github.com/nytimes/react-prosemirror/assets/2272019/869da51d-f526-462b-b79b-47dbd4b0f9d2

### DIFF
--- a/demo/Menu.tsx
+++ b/demo/Menu.tsx
@@ -1,7 +1,7 @@
-import React, { ReactNode } from "react";
+import { toggleMark } from "prosemirror-commands";
 import type { MarkType } from "prosemirror-model";
 import type { EditorState, Transaction } from "prosemirror-state";
-import { toggleMark } from "prosemirror-commands";
+import React, { ReactNode } from "react";
 
 function isMarkActive(mark: MarkType, state: EditorState): boolean {
   if (!state.selection) return false;
--- a/demo/Menu.tsx
+++ b/demo/Menu.tsx
@@ -21,7 +21,6 @@ export function Button(props: {
   title: string;
   onClick: () => void;
 }) {
-
   return (
     <button
       type="button"
--- a/demo/Menu.tsx
+++ b/demo/Menu.tsx
@@ -1,6 +1,6 @@
 import { toggleMark } from "prosemirror-commands";
 import type { MarkType } from "prosemirror-model";
-import type { EditorState, Transaction } from "prosemirror-state";
+import type { EditorState } from "prosemirror-state";
 import React, { ReactNode } from "react";
 
 import { useEditorEventCallback, useEditorState } from "../src/index.js";
--- a/demo/Menu.tsx
+++ b/demo/Menu.tsx
@@ -40,17 +40,15 @@ export function Button(props: {
 export default function Menu() {
   const state = useEditorState();
 
-  const { marks } = state.schema;
-
   const toggleBold = useEditorEventCallback((view) => {
     if (!view) return;
-    const toggleBoldMark = toggleMark(marks["bold"]);
+    const toggleBoldMark = toggleMark(view.state.schema.marks["bold"]);
     toggleBoldMark(view.state, view.dispatch, view);
   });
 
   const toggleItalic = useEditorEventCallback((view) => {
     if (!view) return;
-    const toggleBoldMark = toggleMark(marks["em"]);
+    const toggleBoldMark = toggleMark(view.state.schema.marks["em"]);
     toggleBoldMark(view.state, view.dispatch, view);
   });
 
@@ -59,7 +57,7 @@ export default function Menu() {
       <Button
         className="bold"
         title="Bold (⌘b)"
-        isActive={isMarkActive(marks["strong"], state)}
+        isActive={!!state && isMarkActive(state.schema.marks["strong"], state)}
         onClick={toggleBold}
       >
         B
@@ -67,7 +65,7 @@ export default function Menu() {
       <Button
         className="italic"
         title="Italic (⌘i)"
-        isActive={isMarkActive(marks["em"], state)}
+        isActive={!!state && isMarkActive(state.schema.marks["em"], state)}
         onClick={toggleItalic}
       >
         I
--- a/demo/Menu.tsx
+++ b/demo/Menu.tsx
@@ -42,7 +42,7 @@ export default function Menu() {
 
   const toggleBold = useEditorEventCallback((view) => {
     if (!view) return;
-    const toggleBoldMark = toggleMark(view.state.schema.marks["bold"]);
+    const toggleBoldMark = toggleMark(view.state.schema.marks["strong"]);
     toggleBoldMark(view.state, view.dispatch, view);
   });
 
--- a/demo/Menu.tsx
+++ b/demo/Menu.tsx
@@ -1,0 +1,61 @@
+import React, { ReactNode } from "react";
+import type { MarkType } from "prosemirror-model";
+import type { EditorState, Transaction } from "prosemirror-state";
+import { toggleMark } from "prosemirror-commands";
+
+function isMarkActive(mark: MarkType, state: EditorState): boolean {
+  if (!state.selection) return false;
+  const { from, $from, to, empty } = state.selection;
+  return empty
+    ? !!mark.isInSet(state.storedMarks || $from.marks())
+    : state.doc.rangeHasMark(from, to, mark);
+}
+
+export function Button(props: {
+  children?: ReactNode;
+  isActive: boolean;
+  className?: string;
+  onClick: () => void;
+}) {
+  function handleMouseDown(e: React.MouseEvent) {
+    e.preventDefault();
+    props.onClick();
+  }
+  return (
+    <button
+      className={`button ${props.className} ${props.isActive ? "active" : ""}`}
+      onMouseDown={handleMouseDown}
+    >
+      {props.children}
+    </button>
+  );
+}
+
+export default function Menu(props: {
+  state: EditorState;
+  dispatchTransaction: (transaction: Transaction) => void;
+}) {
+  const { marks } = props.state.schema;
+  return (
+    <div className="menu">
+      <Button
+        className="bold"
+        isActive={isMarkActive(marks["strong"], props.state)}
+        onClick={() =>
+          toggleMark(marks["strong"])(props.state, props.dispatchTransaction)
+        }
+      >
+        B
+      </Button>
+      <Button
+        className="italic"
+        isActive={isMarkActive(marks["em"], props.state)}
+        onClick={() =>
+          toggleMark(marks["em"])(props.state, props.dispatchTransaction)
+        }
+      >
+        I
+      </Button>
+    </div>
+  );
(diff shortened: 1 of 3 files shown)


smoores-dev:
Ah! Cool. Actually, here's what I would suggest instead:

- Move the `Menu` component to be a child prop to `ProseMirror`, alongside the mount div.
- Use the `useEditorState` hook to access the EditorState, rather than passing it as a prop

coolov:
Thank you @smoores-dev. Is this to take advantage of React Context and avoid prop-drilling, or are there other reasons? I would still need to pass down the dispatch transaction function, right?

smoores-dev:
In the case of the EditorState specifically, it's mostly a convenience to avoid prop drilling. But it's important that accessing any other part of the EditorView (including dispatch) happens via the appropriate hooks, or users may experience state tearing. So instead of passing dispatchTransaction as a prop, this should actually use the [`useEditorEventCallback` hook to call `view.dispatch`](https://github.com/nytimes/react-prosemirror#useeditoreventcallback)

coolov:
Got it! Thanks for the quick response.

This actually gets right at something that I have been trying to wrap my head around.

```ts
export function ProseMirrorEditor() {
  const [mount, setMount] = useState<HTMLElement | null>(null);

  // Can I read/update this or is it only for <ProseMirror> to use?
  // ⬇️ ⬇️ ⬇️ ⬇️
  const [editorState, setEditorState] = useState(
    EditorState.create({ schema })
  );

  return (
    <ProseMirror
      mount={mount}
      state={editorState}
      dispatchTransaction={(tr) => {
        setEditorState((s) => s.apply(tr));
      }}
    >
      <div ref={setMount} />
    </ProseMirror>
  );
}
```

If the React managed EditorState cannot be safely accessed/updated outside the react-prosemirror hooks, should it then be exposed to the user?

tilgovi:
Your code doesn't depend on the `EditorView`. The `toggleMark` command doesn't use the view, but other commands do. Using the state outside hooks isn't a problem, but using the `state` attribute of the `EditorView` outside of the hooks is.

I'd probably suggest consuming state and dispatch through the hooks, though, as an idiomatic way to access these that transitions cleanly to using the view if it becomes necessary.

tilgovi:
Actually, there's another reason to use the hooks, too. The `useEditorView` hook (used under the hood by `<ProseMirror>`) does is wrap your `dispatchTransaction` in a call to `ReactDOM.flushSync()`.

Transactions are based on existing state, but ProseMirror commands accept the current state as an argument. There's no equivalent to `setState((current) => next)`. To ensure that transactions are always based on the latest state, every dispatch forces React to flush the update synchronously.

tilgovi:
You bring up a good point about whether we could hide the state and dispatch from users entirely. We deliberately hide the `EditorView` instance, but hiding the state and dispatch functions isn't possible for users who want to keep them outside the view, like in Redux, possibly to run transactions without the view even mounted.

coolov:
@tilgovi thank you! 

> Actually, there's another reason to use the hooks, too. The useEditorView hook (used under the hood by <ProseMirror>) does is wrap your dispatchTransaction in a call to ReactDOM.flushSync().

This is super helpful! I'm still learning about all the mechanisms for preventing "state tearing", I appreciate the pointer. 

> I'd probably suggest consuming state and dispatch through the hooks, though, as an idiomatic way to access these that transitions cleanly to using the view if it becomes necessary.

So, in other words: The code here is safe, but not idiomatic?

Is there an idiomatic way to synchronize React state "sideways", that is: can we "lift state up" so that components that are _not_ descendants of <ProseMirror> can read/set EditorState? Or is this an anti-pattern? Should all components that are ProseMirror aware be children of ProseMirror?

tilgovi:
> Is there an idiomatic way to synchronize React state "sideways", that is: can we "lift state up" so that components that are not descendants of can read/set EditorState? Or is this an anti-pattern? Should all components that are ProseMirror aware be children of ProseMirror?

Pass it however you like to pass state in React. You can use context or props. But if your components are outside the `<ProseMirror>` tree, then you won't have access to the view. That's fine!

I might revise @smoores-dev's comments a little bit. You could leave `<Menu>` outside `<ProseMirror>` because it doesn't depend on the view. That is an idiom. We probably should have other examples in the demo that show how to put things inside the view, though!

tilgovi:
What you've got here is totally fine. It just doesn't showcase the more interesting bits of the library.

smoores-dev:
Right, I think this all works, but if we're going to check it in as part of the demo, I think we should probably use the hooks, just so that we're pointing folks in the right direction!

coolov:
thanks for all the feedback. i updated the example to use the hooks!

demo/Menu.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const toggleBold = useEditorEventCallback((view) => {
+    if (!view) return;
+    const toggleBoldMark = toggleMark(marks["bold"]);
+    toggleBoldMark(view.state, view.dispatch, view);
+  });
+
+  const toggleItalic = useEditorEventCallback((view) => {
+    if (!view) return;
+    const toggleBoldMark = toggleMark(marks["em"]);
```
tilgovi, line 48: nit: you could memoize this outside the callback, but it's not a big deal if you don't. You can even do it at module scope if you know your schema ahead of time.

demo/Menu.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { marks } = state.schema;
+
+  const toggleBold = useEditorEventCallback((view) => {
+    if (!view) return;
+    const toggleBoldMark = toggleMark(marks["bold"]);
+    toggleBoldMark(view.state, view.dispatch, view);
+  });
+
+  const toggleItalic = useEditorEventCallback((view) => {
+    if (!view) return;
```
coolov, line 47: i'm trying to better understand the react-prosemirror api's. when would view be `null` here? and is the expectation to simply drop the event?

demo/Menu.tsx
```diff
@@ -0,0 +1,70 @@
...
+import type { MarkType } from "prosemirror-model";
+import type { EditorState } from "prosemirror-state";
+import React, { ReactNode } from "react";
+
+import { useEditorEventCallback, useEditorState } from "../src/index.js";
+
+function isMarkActive(mark: MarkType, state: EditorState): boolean {
+  if (!state.selection) return false;
+  const { from, $from, to, empty } = state.selection;
+  return empty
```
tilgovi, line 11: I wonder if you need the ternary or if this would work, regardless of selection emptiness: `mark.isInSet(state.storedMarks) || state.doc.rangeHasMark(from, to, mark)`.

It's possible that changing the selection clears the stored marks anyway.

demo/Menu.tsx
smoores-dev: On the first render pass, the view will be null, because it's constructed in a layout effect. That means that we sort of have to type the context as returning an optional view, even though in practice, I think it would always be defined by the time the callback actually runs.

demo/Menu.tsx
coolov: the code is lifted from the prosemirror-example-setup (added a comment):
https://github.com/ProseMirror/prosemirror-example-setup/blob/master/src/menu.ts#L58

it has the same condition, but i don't know why 🤷

tilgovi: Cool.

demo/Menu.tsx
tilgovi: Yeah, it's expected that you can just drop the event in this case.

coolov: thanks for explaining. just throwing it out there: have we considered not firing the callback if `view` is null? this way we could make `view` non-nullable which would make the userland code a little less verbose.

tilgovi: I don't think we had.

coolov: I'd be happy to make a PR if it makes sense!

demo/Menu.tsx
```diff
@@ -0,0 +1,72 @@
...
+      className={`button ${props.className} ${props.isActive ? "active" : ""}`}
+      onMouseDown={handleMouseDown}
+    >
+      {props.children}
+    </button>
+  );
+}
+
+export default function Menu() {
+  const state = useEditorState();
```
coolov, line 39: just realizing that this is nullable also, i'm guessing for similar reason as [below](https://github.com/nytimes/react-prosemirror/pull/81/files#r1466748866)? not sure why it passed the typechecker, will follow up...

demo/Menu.tsx
coolov: Actually, after poking around I realize that my proposal is somewhat limited. 

> On the first render pass, the view will be null, because it's constructed in a layout effect. 

I'm reading [useEditorView](https://github.com/nytimes/react-prosemirror/blob/main/src/hooks/useEditorView.ts) and [ProseMirrorInner](https://github.com/nytimes/react-prosemirror/blob/main/src/components/ProseMirrorInner.tsx) and learning that the EditorContext value has [nullable props](https://github.com/nytimes/react-prosemirror/blob/main/src/contexts/EditorContext.ts) for both EditorState and EditorView. From what I gather it seems like we are waiting for the editor to mount before [we initialize these values](https://github.com/nytimes/react-prosemirror/blob/main/src/hooks/useEditorView.ts#L105).

Is there something we can do here to guarantee these initial values? Defer [the rendering of the provider and its children](https://github.com/nytimes/react-prosemirror/blob/main/src/components/ProseMirrorInner.tsx#L68)?

It would be wonderful to not have to null-check the return values of all the hooks because of this initial-render case.

tilgovi: We can't defer rendering the children because we need the mount point to render so that we can mount the view. It's the tradeoff we made to give the user full control of nesting whatever components they want under `<ProseMirror>` and styling the editing root however they want. We have to actually render the element that sets the mount ref.

tilgovi: So, I think we need the context to be nullable values, but we maybe don't need to invoke any callbacks for `useEditorEvent` or `useEditorEffect` unless the view is defined.

tilgovi: We can open an issue to discuss! It'd be a non-breaking change to make it always defined, but it'd be a breaking change to have to revert that. So, let's talk about it and be sure!

coolov: Got it! 

I already have the change on a branch and can open a PR where we can discuss.

I'll follow up elsewhere regarding `useEditorState();` returning a nullable value.

tilgovi: 🚀

demo/Menu.tsx
coolov: https://github.com/nytimes/react-prosemirror/pull/84

jrapala:
Pushed a couple of small accessibility improvements to the demo

PR #83, opened 2024-01-25
Makes EditorView non-nullable in `useLayoutGroupEffect` and `useEditorEventCallback`

coolov:
## What?

Currently the `view` prop in `useLayoutGroupEffect` and `useEditorEventCallback` is nullable, this means that we need to assert that the value exists before using it:

```js
const onClick = useEditorEventCallback((view) => {
  if (view) {
    // do things
  }
});
```

By making the value non-nullable we no longer need this check!

## How?

This PR proposes that we only fire the  callbacks in the effects if `view` is set. This way we can guarantee that the value is never `null` and always of type `EditorView`.

So why is the value sometimes `null` in the first place? This has to do with how the initial render bootstraps the [EditorContext](https://github.com/nytimes/react-prosemirror/blob/main/src/contexts/EditorContext.ts) that holds the EditorView. See @smoores-dev [comment in this thread](https://github.com/nytimes/react-prosemirror/pull/81#discussion_r1466748866):

> On the first render pass, the view will be null, because it's constructed in a layout effect. That means that we sort of have to type the context as returning an optional view, even though in practice, I think it would always be defined by the time the callback actually runs.

For more details about how the creation of the EditorContext is deferred, see [useEditorView](https://github.com/nytimes/react-prosemirror/blob/main/src/hooks/useEditorView.ts#L105)

### DIFF
--- a/src/hooks/useEditorEffect.ts
+++ b/src/hooks/useEditorEffect.ts
@@ -34,9 +34,7 @@ export function useEditorEffect(
   // be defined inline and run on every re-render.
   useLayoutGroupEffect(
     () => {
-      if (editorView) {
-        effect(editorView);
-      }
+      if (editorView) effect(editorView);
     },
     // The rules of hooks want to be able to statically
     // verify the dependencies for the effect, but this will
--- a/src/hooks/useEditorEffect.ts
+++ b/src/hooks/useEditorEffect.ts
@@ -20,7 +20,7 @@ import { useLayoutGroupEffect } from "../contexts/LayoutGroup.js";
  * EditorView lives in an ancestor component.
  */
 export function useEditorEffect(
-  effect: (editorView: EditorView | null) => void | (() => void),
+  effect: (editorView: EditorView) => void | (() => void),
   dependencies?: DependencyList
 ) {
   const { editorView } = useContext(EditorContext);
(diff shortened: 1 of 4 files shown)


smoores-dev:
Awesome, thanks @coolov. I actually think this is a pretty big win; more than one user has expressed confusion or annoyance that the `view` might be `null`.

This feels good to me, though as @tilgovi pointed out in the other thread, undoing this would be a breaking change, so it's worth thinking about critically. @tilgovi, did you have any other thoughts you wanted to raise on the topic?

src/hooks/useEditorEffect.ts
```diff
@@ -33,7 +33,9 @@ export function useEditorEffect(
   // every time it changes, because it will most likely
   // be defined inline and run on every re-render.
   useLayoutGroupEffect(
-    () => effect(editorView),
+    () => {
+      if (editorView) effect(editorView);
```
coolov, line 37: we currently assume that this is never false, but what if it sometimes is for some unforeseen reason, what do we do? should we fail silently as we do here, warn, or throw?

smoores-dev: I think the answer to this is no, it's not a failure if `editorView` is undefined when this runs, it just means it's not ready to run the effect yet

smoores-dev (approved):

src/hooks/useEditorEffect.ts
coolov: yeah that aligns with my thinking. great!

Issue #82, opened 2024-01-25
Can we make the return value of useEditorState non-nullable? 


coolov:
`useEditorState` is "nullable" even if `EditorState` is guaranteed to have been initialized before it's usage.

Example:

```ts
function ProseMirrorEditor() {
  const [mount, setMount] = useState(null);

  // the initial editor state is created
  const [editorState, setEditorState] = useState(
    EditorState.create({ schema })
  );

  return (
    <ProseMirror
      mount={mount}
      state={editorState}
      dispatchTransaction={(tr) => setEditorState((s) => s.apply(tr))}
    >
      <div ref={setMount} />
      <ChildA state={state}/>
      <ChildB/>
    </ProseMirror>
  );
}

// EditorState is guaranteed to exist when it is passed down as a prop
function ChildA(props: { state: EditorState }) {
   // state is always of type EditorState, we can access it without null checking
   console.log(state.schema);
   
   // ... //
}

// EditorState is nullable when it is accessed via a hook
function ChildB() {
  const state = useEditorState();

  // The type of state is either null or EditorState, so we need to null check before using it
  if (state) {
    console.log(state.schema);
  }
  
  // ... //
}
```

My understanding, based [on this conversation](https://github.com/nytimes/react-prosemirror/pull/81#discussion_r1466748866) is that this is because of how we bootstrap [the EditorView](https://github.com/nytimes/react-prosemirror/blob/main/src/hooks/useEditorView.ts) and the EditorContext. 

Is there a way to change this behavior so that:
1) The state value in the EditorContext is created immediately/synchronously so that it can never be `null`?
2) Can we do this safely without risking "state tearing"? (If we can't, why?)
3) Make the `useEditorState` hook always return `EditorState` and never `null`

PR #84, opened 2024-01-26
Makes the return value of useEditorState non-nullable

coolov:
Fixes https://github.com/nytimes/react-prosemirror/issues/82

I left a comment on the one-line change below. Am I missing something crucial here?

### DIFF
--- a/.yarn/versions/53dfe063.yml
+++ b/.yarn/versions/53dfe063.yml
@@ -1,0 +1,2 @@
+undecided:
+  - "@nytimes/react-prosemirror"
--- a/.yarn/versions/53dfe063.yml
+++ b/.yarn/versions/53dfe063.yml
@@ -1,2 +1,2 @@
 releases:
-  "@nytimes/react-prosemirror": patch
+  "@nytimes/react-prosemirror": minor
--- a/.yarn/versions/53dfe063.yml
+++ b/.yarn/versions/53dfe063.yml
@@ -1,2 +1,2 @@
-undecided:
-  - "@nytimes/react-prosemirror"
+releases:
+  "@nytimes/react-prosemirror": patch
--- a/src/components/ProseMirrorInner.tsx
+++ b/src/components/ProseMirrorInner.tsx
@@ -53,7 +53,9 @@ export function ProseMirrorInner({
   });
 
   const editorState =
-    "state" in editorProps ? editorProps.state : editorView?.state ?? null;
+    "defaultState" in editorProps
+      ? editorProps.defaultState
+      : editorProps.state;
 
   const editorContextValue = useMemo(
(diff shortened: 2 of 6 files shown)


src/components/ProseMirrorInner.tsx
```diff
@@ -53,7 +53,9 @@ export function ProseMirrorInner({
   });
 
   const editorState =
-    "state" in editorProps ? editorProps.state : editorView?.state ?? null;
+    "defaultState" in editorProps
+      ? editorProps.defaultState
+      : editorProps.state;
```
coolov, line 58: The fix is here: 
`<ProseMirror>` requires either `state` or `defaultState` to be passed in as props. If we choose one or the other the editorState value will never be `null`.

smoores-dev, line 58: This doesn't look correct to me. This `editorState` variable should be the current editorState from the view (note how before it was being read from `editorView.state` if there was no state prop).

If it's set to the `defaultState`, then it will never update for users that pass a `defaultState`!

coolov: I had a feeling this was too easy to be true!

smoores-dev: I _think_ that, instead of this, it would be technically correct to do:

```ts
"defaultState" in editorProps
     // Only use the default state as a fallback for the first render where `editorView` isn't initialized yet
      ? (editorView?.state ?? editorProps.defaultState)
      : editorProps.state;
```

coolov: so... how does the useState hook react to the state reference that is referenced from `editorView.state`? is it tracked in react state somewhere?

smoores-dev: This is why there's a forceUpdate call in the useEditorView hook!

coolov: ah, that makes sense. gonna try your suggestion above!

coolov: yeah, that seems to work!

.yarn/versions/53dfe063.yml
```diff
@@ -0,0 +1,2 @@
+releases:
+  "@nytimes/react-prosemirror": patch
```
smoores-dev, line 2: Should we make this a minor version change, like #83? It's also "not a breaking change", but folks may have strict type checking or linting rules that error with unnecessary null checks, which would result in broken pipelines after incorporating these changes

coolov: that's a great suggestion. will update!

smoores-dev (approved):

tilgovi (approved):

PR #88, opened 2024-01-30
Remove unnecessary null guarding in the demo menu

coolov:
A couple of changes to the demo menu (https://github.com/nytimes/react-prosemirror/pull/81) that removes unnecessary null guarding after incorporating the changes from 
https://github.com/nytimes/react-prosemirror/pull/84 and https://github.com/nytimes/react-prosemirror/pull/83


smoores-dev:
So much nicer!

smoores-dev (approved):

PR #90, opened 2024-01-30
Remove duplicate plugin file and update useComponentEventListeners block comment

ariellebryn:
Seems like the `componentEventListenersPlugin` file remained when trying to rename it to `componentEventListeners` in #33. Only the latter is used.

Similarly, `reactEventsPlugins` was renamed to `componentEventListenersPlugin` in #17, but this one comment wasn't updated.


coolov (approved):

smoores-dev:
@ariellebryn you can choose decline for version bump (that should be one of the options for `yarn version check -i`)!

PR #91, opened 2024-01-31
Map positions forward to maintain key stability

### LINKED ISSUE #85: Caret is stuck at the start of node after applying input rule
During the migration to the ReactEditorView, I noticed some caret issues in conjunction with input rules.

I added a minimal reproduction in [this commit](https://github.com/konstantinmuenster/react-prosemirror/commit/2005c1a0c0eb0fc607f8a35f96f352a34d0cf262).

**Steps to reproduce**
1. Turn a paragraph into a list via input rule
2. Select all text
3. Start writing again
4. -> the caret sticks at the start of node

I haven't looked into it that much since I had hoped you have a better gut feeling of what could be the issue.

https://github.com/nytimes/react-prosemirror/assets/46243719/a981b434-2e24-4e7e-aaa4-ba59babbe218

smoores-dev:
Rather than attempting to map the new node positions backward to determine what the key used to be for that node, start with the previous positions and map them forward, taking care to skip any positions that are deleted by the transaction, and adding any new nodes that didn't exist in the old doc. This improves stability and correctness in more complex cases, such as splitting or wrapping nodes, and nested deletions.

Fixes #85.


smoores-dev:
We should also patch this into the main branch, in place of #74 !

PR #93, opened 2024-02-06
Update props atomically and consistently

tilgovi:
Whether the editor state is controlled or uncontrolled, update all props atomically and only in a layout effect after rendering.

- Put uncontrolled state into a state hook. The default dispatch can use this hook to update the editor state so that view updates always occur in a layout effect after a render cycle, even with uncontrolled state.

- Type the components and hooks to account for passing both defaultState and state. Handle this in the same way that React itself does, writing a warning to the console on the first occurrence, such that users that do not use TypeScript still get some guidance.

- Merge the layout effects that update the view so that all of the props of the view get updated simultaneously. Otherwise, any view props that may close over external state, such as decorations, may be out of date during the update of non-state props, which can cause errors.

- Push all logic from the inner ProseMirror component into the hook so that all props beyond the base ProseMirror editor props get handled in one place.

- Inline the inner component into the main ProseMirror component module. Name the component simply "Editor" to align with its purpose of owning the editor view and providing the editor context.

- Give the contexts display names so that their purpose is clearer when viewing them in React development tools.

- Use "state" in all examples, instead of "editorState", for simplicity. This library is not Redux and software developers are used to handling naming collisions, if they occur. Redux users are used to seeing React examples that use the state hook and knowing that it is different from Redux state. ProseMirror state is similar in this way.


src/hooks/useEditorView.ts
```diff
@@ -59,70 +36,86 @@ function withFlushedDispatch(
...
 export function useEditorView<T extends HTMLElement = HTMLElement>(
   mount: T | null,
-  props: EditorProps
-): EditorView | null {
-  const [view, setView] = useState<EditorView | null>(null);
-  const forceUpdate = useForceUpdate();
-
-  const editorProps = withFlushedDispatch(props, forceUpdate);
+  props: Props
+): EditorContextValue {
```
tilgovi, line 44: The new return value here is a breaking change to the hook, so I've made this a minor version bump. I don't expect this will be troubling to anyone. Every use we have internally is through the ProseMirror component, and not through the hook directly.

tilgovi:
Converting to a draft while I add a test.

src/hooks/useEditorView.ts
```diff
@@ -59,70 +36,86 @@ function withFlushedDispatch(
...
+  if (process.env.NODE_ENV !== "production") {
+    if (
+      props.defaultState !== undefined &&
+      props.state !== undefined &&
+      !didWarnValueDefaultValue
+    ) {
+      console.error(
+        "A component contains a ProseMirror editor with both value and defaultValue props. " +
+          "ProseMirror editors must be either controlled or uncontrolled " +
+          "(specify either the value prop, or the defaultValue prop, but not both). " +
```
smoores-dev, line 54:
```suggestion
          "(specify either the state prop, or the defaultState prop, but not both). " +
```

tilgovi:
Okay, ready for review. I thought I wasn't going to have time tonight, so I pushed it as a draft, but it turned out to be easy to write the test.

src/hooks/useEditorView.ts
```diff
@@ -59,70 +36,86 @@ function withFlushedDispatch(
...
   mount: T | null,
-  props: EditorProps
-): EditorView | null {
-  const [view, setView] = useState<EditorView | null>(null);
-  const forceUpdate = useForceUpdate();
-
-  const editorProps = withFlushedDispatch(props, forceUpdate);
+  props: Props
+): EditorContextValue {
+  if (process.env.NODE_ENV !== "production") {
```
smoores-dev, line 45: Is this a safe check? I feel like not all transpilers will polyfill this

tilgovi:
@ariellebryn tagging you here because you asked about why updates to React components and the editor view seemed to happen in a different order depending on whether we used controlled state, and this PR should fix that.

src/hooks/useEditorView.ts
```diff
@@ -59,70 +36,86 @@ function withFlushedDispatch(
...
-    return () => {
-      if (view) {
-        view.destroy();
+  function dispatchTransaction(this: EditorView, tr: Transaction) {
+    flushSync(() => {
+      if (props.dispatchTransaction) {
+        props.dispatchTransaction.call(this, tr);
+      } else {
+        setState((s) => s.apply(tr));
       }
```
smoores-dev, line 84: If I'm reading this correctly, it's making the assumption that users won't pass `dispatchTransaction` and `defaultState`, but that is a supported use case now, and I think it should continue to be. React allows this, for example, for uncontrolled components; it allows the consumer to execute side effects when changes occur, even if they haven't lifted the state.

tilgovi:
@eelzon tagging you here because we talked about how I thought the non-atomicity was a holdover from a performance workaround from before prosemirror-view got smarter about not redrawing node views. It turns out it was actually something we introduced as a regression here when adding support for uncontrolled state.

I tried to find less invasive changes that felt logical, but it felt like the smoothest thing was to just get rid of almost everything in `<ProseMirrorInner />` altogether, which necessitated the hook returning the whole context. Every other way I tried to do this the props got messy and the division of responsibilities between `<ProseMirrorInner />` and the hook felt awkward.

src/components/ProseMirror.tsx
```diff
@@ -1,8 +1,25 @@
...
 
-import { ProseMirrorInner, ProseMirrorProps } from "./ProseMirrorInner.js";
+interface Props extends EditorProps {
+  mount: HTMLElement | null;
+  children?: ReactNode | null;
+  defaultState?: EditorState;
+  state?: EditorState;
+  plugins?: readonly Plugin[];
+  dispatchTransaction?(this: EditorView, tr: Transaction): void;
+}
```
tilgovi, line 17: This is repeated, rather than imported. I'd like to leave it that way for the moment because I think maybe we should consider exporting only the `<ProseMirror />` component, and hiding the hook, like we do for `<LayoutGroup />`.

src/components/__tests__/ProseMirror.test.tsx
```diff
@@ -14,6 +13,12 @@ import {
 } from "../../testing/setupProseMirrorView.js";
 import { ProseMirror } from "../ProseMirror.js";
 
+// Mock `ReactDOM.flushSync` to call `act` to flush updates from DOM mutations.
+jest.mock("react-dom", () => ({
+  ...jest.requireActual("react-dom"),
+  flushSync: (fn: () => void) => act(fn),
+}));
+
```
smoores-dev, line 21: ❤️

src/hooks/useEditorView.ts
```diff
@@ -1,56 +1,33 @@
...
 
-function withFlushedUpdates<This, T extends unknown[]>(
-  fn: (this: This, ...args: T) => void
-): (...args: T) => void {
-  return function (this: This, ...args: T) {
-    flushSync(() => {
-      fn.call(this, ...args);
-    });
-  };
-}
```
tilgovi, line 17: I think this file is easier to read now without this higher-order function making things more abstract.

src/hooks/useEditorView.ts
```diff
@@ -1,56 +1,33 @@
...
+  },
+});
 
-type EditorStateProps =
-  | {
-      state: EditorState;
-    }
-  | {
-      defaultState: EditorState;
-    };
```
tilgovi, line 29: I think it's also easier to follow the logic of warning when `defaultValue` and `value` are both specified than it is to follow conditional type narrowing.

src/components/ProseMirror.tsx
smoores-dev: I think I agree about hiding the hook. I think we should still be able to import the props from the hook, though? I guess I don't see why hiding the hook would entail not being able to share the props, but I suspect I'm missing something.

src/hooks/useEditorView.ts
```diff
@@ -1,56 +1,33 @@
...
-        tr: Transaction
-      ) {
-        const flushedDispatch = withFlushedUpdates(
-          props.dispatchTransaction ?? defaultDispatchTransaction
-        );
-        flushedDispatch.call(this, tr);
-        if (!("state" in props)) forceUpdate();
-      },
-    },
-  };
```
tilgovi, line 53: It's easier to write this to update state through React by inlining it in the component. It provides a nice narrative structure wherein the component handles each of the direct editor props in turn (`state`, `plugins`, and `dispatchTransaction`). It's easier to see that the props overall aren't straightforward to memoize without some shallow comparisons, and we are better off delegating that to ProseMirror itself, I think.

src/hooks/useEditorView.ts
smoores-dev: I agree, but I'm sad about the loss of DX. Would we be open to keeping this type for the public props, and casting it internally to make it simpler for us?

src/components/ProseMirror.tsx
tilgovi: I think what I'm seeing is possibly just moving it all to one file next.

src/hooks/useEditorView.ts
tilgovi: Do you think it's really bad? Folks will still get autocomplete telling them that these both exist, and I think they're not likely to try to specify both.

tilgovi: This matches what React itself does with `<input />`, if that's any reassurance.

src/hooks/useEditorView.ts
tilgovi: Most bundlers will. It's true not everything will, by default, but it gives me some assurance than React itself does this in its entrypoint. So, any project set up to import React should probably be set up to do this.

src/hooks/useEditorView.ts
smoores-dev: I don't think it's _really_ bad, for sure. I just think it's very nice when you can specify precisely the type you mean, especially for a third-party lib, to remove any potential for confusion, and since we already had it, I'm sad to see it go

src/hooks/useEditorView.ts
tilgovi: Yup. I think we could maybe make a deviation from EditorView here and stop passing the view as the `this` context so that it's not tempting to write `this.updateState()` in the transaction handler.

You can observe transactions, if you want to for some reason, but unless you control the state you shouldn't actually see the state in that callback or set it.

src/hooks/useEditorView.ts
smoores-dev: I'm worried that React does this because React bundles its own source code, and has a specific dev build that is distinct from the prd build. Does the compiled React bundle actually still say process.env?

src/hooks/useEditorView.ts
tilgovi: I think it'll be messy to try to have it both ways. I value the readability here and doubt there's much of any DX hit to users.

tilgovi: I'll try once more, or feel free to propose something.

src/hooks/useEditorView.ts
```diff
@@ -59,70 +36,86 @@ function withFlushedDispatch(
...
   useLayoutEffect(() => {
-    if (stateProp) view?.setProps({ state: stateProp });
-  }, [view, stateProp]);
-
-  return view;
+    if (view) {
+      if (view.dom === mount) {
+        view.setProps(directEditorProps);
+      } else {
+        view.destroy();
```
saranrapjs, line 104: In the previous implementation, we destroy the view in the `useLayoutEffect` cleanup function. Trying to understand this: this implementation is functionally equivalent, because if the `mount` effect argument changes, this should have the same effect of unmounting (and subsequently, remounting with the logic below) by calling `view.destroy()` at the right moment?
It might be nice to have a test for unmounting.

src/hooks/useEditorView.ts
tilgovi: React ships two different versions, but the entrypoint `index.js` requires one or the other based on this exact condition.

Where we really wouldn't want this is if we had a UMD build that people were expected to put in their browsers directly, without a bundler. But I don't think that's a realistic use for us.

tilgovi: If we do ever have such a thing, we could generate two builds with `if (__DEV__)` like React does, for the UMD use case.

src/hooks/useEditorView.ts
smoores-dev: I'm not sure I understand why. I mean I do think calling `this.updateState()` from a `dispatchTransaction` prop is goofy and maybe worth avoiding (though I think it was handled more or less fine with the `forceUpdate()`), but `<input />` `onChange` handlers get an event that allows you retrieve the entire full state of the element.

src/hooks/useEditorView.ts
smoores-dev: Got it! Thanks for explaining ❤️

smoores-dev:
This is a very good catch @tilgovi, thank you for addressing this!!!

src/hooks/useEditorView.ts
smoores-dev: Would you feel alright about pulling this out into a separate change? I think I would like to think about and discuss it more, but don't want to block this fix

src/hooks/useEditorView.ts
tilgovi: Yeah, good call out.

Setting the state within an effect triggers an immediate re-render, and since this is happening at an ancestor of any component that could access the view it should be the case that they've all run their effects already. So, the net result either way is an immediate re-render of the descendants with a null view.

I guess this way we'll destroy the node views first, which could cause portals to unregister, thereby forcing an update of an even higher component first.

I think it's worth backing this change out of this PR and considering this part more carefully. Thanks.

src/hooks/useEditorView.ts
tilgovi: What do you want me to pull out? I haven't changed the signature of `dispatchTransaction` here, I'm just suggesting we consider it.

tilgovi:
> This is a very good catch @tilgovi, thank you for addressing this!!!

Yay! We actually cause this because we started closing over the decorations from Redux in a refactor to move Oak onto the open source package. Paying dividends!

src/hooks/useEditorView.ts
smoores-dev: Ah, this is more subtle than I remembered/was envisioning. Sorry, I could have been being clearer about this the whole time. The `forceUpdate()` from before would handle the case where someone called `this.updateState()` from `dispatchTransaction`. Storing the editor state in React state here when the component is uncontrolled is definitely the right move, but I think if there's a dispatchTransaction _and_ a defaultState right now, we wouldn't re-render the component tree. So what I'm asking to be backed out is that regression!

tilgovi: Oh, I see! I have to think on that. I think that's pretty central to this PR, so rather than back out I think we need to address it.

Handling this correctly might really require intercepting calls to `updateState` or `setProps`. I dunno.

I'm almost tempted to just rip out support for uncontrolled state at this point.

smoores-dev: > I think that's pretty central to this PR

I don't understand why that would be; couldn't we do a forceUpdate() after calling `props.dispatchTransaction()` if `props.state` is undefined?

tilgovi: I guess we could force an update, but as long as we're storing the uncontrolled state in a state hook, there's no point. That state hasn't changed. If there's other reactive state that the use changed with their `dispatchTransaction`, then that'll cause a render.

I could see calling `setState` before calling the user's `dispatchTransaction` so that, much like regular inputs, the view instance already has an updated value at that point and, like `onChange` you could observe it. In the case of `dispatchTransaction` that's through `this.state`. In the case of `onChange`, `event.target.value`.

I don't know that we should do it when state is controlled, unless we want folks to just observe `this.state` and set it somewhere in the common case. Internally, in the Oak application, we generally apply the transaction ourselves in Redux reducers, such that applying it in the hook is duplicative work.

If we wanna get fancy, we could call `dispatchTransaction` with some kind of proxy object as `this`.

tilgovi: I updated this to call `setState()` before calling `props.dispatchTransaction()` if `props.state` is not specified.

I think that should cause any re-render triggered by `dispatchTransaction` higher up in the tree to flush the uncontrolled state change to the view, doesn't cause any redundant work for the case of controlled state, and ensures that transactions get applied even when `dispatchTransaction` and `defaultState` get specified together.

If you don't try to control the state, it'll be semi-controlled internally, in the sense that we keep state in the hook. A `dispatchTransaction` prop won't be able to observe the state having been applied to the actual view instance yet, but I think maybe this is good enough.

tilgovi: I've now added a test for this case, too.

tilgovi: Let me know if I captured and addressed your concerns correctly, here.

src/hooks/useEditorView.ts
tilgovi: I've backed that out.

src/hooks/useEditorView.ts
tilgovi: I gave it one more effort, but I think it's awkward. Should passing `state={undefined} defaultState={something}` be allowed by typing? I kinda think yes.

Also, if we type them as a union then checking for user error for the non-TypeScript users becomes awkward. TypeScript tries to tell us that we can't, that that can't happen, but of course it can.

tilgovi: When I say "should passing ... be allowed" what I mean is that while TypeScript will allow you to, say, spread an object with extra props into a component, that might cause `"foo" in props` and `props.foo === undefined` to be different. I think the latter is the appropriate check that we should be making in the implementation.

src/hooks/useEditorView.ts
```diff
@@ -59,31 +36,66 @@ function withFlushedDispatch(
...
+    flushSync(() => {
+      if (!props.state) {
+        setState((s) => s.apply(tr));
+      }
+
+      if (props.dispatchTransaction) {
+        props.dispatchTransaction.call(this, tr);
+      }
+    });
+  }
```
smoores-dev, line 88: This does still have the potentially weird effect of basically overriding a consumers' `updateState()` in `dispatchTransaction()` (since we explicitly set the state on the view later on, in the layout effect), but I think that's probably the most correct outcome (and I would still like to think about warning about it in the future I think).

smoores-dev: I don't think we have to do anything about this, just making note of it!

tilgovi: Yup. I think we might be able to defer this a bit. I'd like to see how things look once we have the pure React view implementation. Maybe then we can actually store the view's state in React or something.

smoores-dev (approved):

src/hooks/useEditorView.ts
tilgovi: You, too. Great questions.

src/hooks/useEditorView.ts
smoores-dev: Yeah this definitely feels less fraught in the react-editor-view branch, since we don't have to use layout effects to update the state in the view!

eelzon (approved):
yesssssssss!

PR #100, opened 2024-02-12
Stop exporting the low-level building blocks of the main API

tilgovi:
Until APIs stabilize and users demonstrate a need for more control, remove the layout group and low-level editor exports.


tilgovi:
The motivation here is to focus the documentation on things people really need and free up some namespace. It's been bugging me that `useEditorView` doesn't do something like `useEditorState` but for accessing the view and it bugs me that the options of `useEditorView`, which aren't quite "props" (because it's not a component) leak out into the public interface.

smoores-dev (approved):

PR #103, opened 2024-02-23
Always call LayoutGroup effect destroy functions.

smoores-dev:
This resolves an issue where LayoutGroup effects were not cleaned up when the LayoutGroup itself was unmounted. React executes layout effect destroy functions from parent to child on unmount, which is opposite how it runs during normal render. The result was that the destroyQueue was never processed on unmount.

Now, if the LayoutGroup is being unmounted, we call destroy functions immediately, without queuing.

Note: This also fixes the Prettier config for VS Code, which got messed up when we upgraded to v3!

### DIFF
--- a/src/components/LayoutGroup.tsx
+++ b/src/components/LayoutGroup.tsx
@@ -18,6 +18,7 @@ export interface LayoutGroupProps {
 export function LayoutGroup({ children }: LayoutGroupProps) {
   const createQueue = useRef(new Set<() => void>()).current;
   const destroyQueue = useRef(new Set<() => void>()).current;
+  const isMounted = useRef(false);
 
   const forceUpdate = useForceUpdate();
   const isUpdatePending = useRef(true);
@@ -42,8 +43,12 @@ export function LayoutGroup({ children }: LayoutGroupProps) {
       return () => {
         createQueue.delete(create);
         if (destroy) {
-          destroyQueue.add(destroy);
-          ensureFlush();
+          if (isMounted.current) {
+            destroyQueue.add(destroy);
+            ensureFlush();
+          } else {
+            destroy();
+          }
         }
       };
     },
@@ -59,6 +64,13 @@ export function LayoutGroup({ children }: LayoutGroupProps) {
       destroyQueue.clear();
     };
   });
+
+  useLayoutEffect(() => {
+    isMounted.current = true;
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
 
   return (
     <LayoutGroupContext.Provider value={register}>
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,6 @@
   },
   "typescript.tsdk": ".yarn/sdks/typescript/lib",
   "typescript.enablePromptUseWorkspaceTsdk": true,
-  "prettier.prettierPath": ".yarn/sdks/prettier/index.js",
   "eslint.nodePath": ".yarn/sdks",
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
@@ -18,5 +17,6 @@
   },
   "[yaml]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "prettier.prettierPath": ".yarn/sdks/prettier/index.js"
 }
--- a/.yarn/sdks/prettier/index.js
+++ b/.yarn/sdks/prettier/index.js
@@ -17,4 +17,4 @@ if (existsSync(absPnpApiPath)) {
 }
 
 // Defer to the real prettier/index.js your application uses
-module.exports = absRequire(`prettier/index.js`);
+module.exports = absRequire(`prettier/index.cjs`);
--- a/.yarn/sdks/prettier/package.json
+++ b/.yarn/sdks/prettier/package.json
@@ -1,6 +1,6 @@
 {
   "name": "prettier",
-  "version": "2.8.3-sdk",
+  "version": "3.2.5",
   "main": "./index.js",
   "type": "commonjs"
 }
--- a/.yarn/versions/8474c081.yml
+++ b/.yarn/versions/8474c081.yml
@@ -1,0 +1,2 @@
+releases:
+  "@nytimes/react-prosemirror": patch


tilgovi:
Don't the deleted lines 57-60 destroy the layout effects?

smoores-dev:
Yes, this now happens on line 43, instead (which is executed as a layout effect cleanup function in the child component)

tilgovi:
I don't have a use case in mind where grouping destroys is actually important, but the asymmetry makes me uncomfortable. It feels more surprising somehow.

I think what's maybe important is that if a descendant of the group unmounts, it should trigger a flush of the group, because the other layout effects of descendants that continue to exist might need to update. Consider a scenario where some other component in the tree has a layout group effect with no dependencies. We want to force the whole group to rerender so that re-runs after the unmount of another component, no?

smoores-dev:
If I'm understanding correctly, we were already trying to do what you're describing with line 46. It's not sufficient, because the parent has already unmounted by the time the trigger to flush runs, and so it does not re-run its cleanup effect.

I agree that this is inconsistent, but:

- I actually don't think it's surprising; the behavior is now that the create effect runs after the view create effect, and the destroy effect runs before it. This actually is what I had been assuming was happening before I looked into this, and I think it makes some intuitive sense
- React itself is inconsistent in exactly this way, so I think we get a pass! Haha

smoores-dev:
There is another approach where we keep the destroyQueue as is, and special-case the scenario where the LayoutGroup is unmounted. This would allow us to actually be _more_ consistent than React is, and _always_ run child effect destroy functions after the LayoutGroup's. Does that seem better?

src/hooks/useEditorEffect.ts
```diff
@@ -34,7 +34,9 @@ export function useEditorEffect(
   // be defined inline and run on every re-render.
   useLayoutGroupEffect(
     () => {
-      if (editorView) effect(editorView);
+      if (editorView) {
+        return effect(editorView);
+      }
     },
```
smoores-dev, line 40: When we added this check for `editorView`, we unintentionally stopped returning the result of calling `effect`. This meant that we were no longer executing cleanup functions ever!

smoores-dev:
Just added a second commit to demonstrate the alternative approach!

tilgovi:
> There is another approach where we keep the destroyQueue as is, and special-case the scenario where the LayoutGroup is unmounted. This would allow us to actually be _more_ consistent than React is, and _always_ run child effect destroy functions after the LayoutGroup's. Does that seem better?

I was just about to suggest something like this, but let me take a bit to digest.

src/components/LayoutGroup.tsx
```diff
@@ -18,6 +18,7 @@ export interface LayoutGroupProps {
 export function LayoutGroup({ children }: LayoutGroupProps) {
   const createQueue = useRef(new Set<() => void>()).current;
   const destroyQueue = useRef(new Set<() => void>()).current;
+  const isMounted = useRef(true);
```
tilgovi, line 21: Is this actually true, initially? 😆

smoores-dev: I meeeaaannnn... I guess it depends on your definition of mounted haha. Given the way that this is being used, I suppose it's safe to set this to false initially and set it true in a useLayoutEffect creator.

tilgovi: If so, I'm tempted by the idea of not queueing child destroys, but providing a simpler explanation for why.

smoores-dev: It's not, I've checked. It is child->parent in regular updates, and actually behaves differently and runs parent->child in during unmounts!

src/hooks/useEditorEffect.ts
smoores-dev: #104

src/components/LayoutGroup.tsx
tilgovi: That's true even if both the parent and the child have effects with dependencies that changed and the render was caused by a state update at the parent or above?

smoores-dev: I was testing with effects with no dependencies because it seemed easier to coordinate. But yes, if you use effects with no dependencies, and update the state at the parent, effect cleanups are run from child -> parent. If you unmount the parent, they are run from parent -> child

tilgovi: That's super wild.

smoores-dev: For anyone that wants to follow along, here's a [minimal reproduction on CodeSandbox](https://codesandbox.io/p/sandbox/goofy-brook-tyc87m?layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clsz0h67t00062e69ppema5fd%2522%252C%2522sizes%2522%253A%255B70%252C30%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clsz0h67t00022e696gdauoqe%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clsz0h67t00032e69d0dlryqu%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clsz0h67t00052e69c281fr1z%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B50%252C50%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clsz0h67t00022e696gdauoqe%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clsz0h67t00012e69sqat2v6j%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252Fsrc%252Findex.js%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%252C%257B%2522id%2522%253A%2522clsz0ieva00022e692uqb170o%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252Fsrc%252FApp.js%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%255D%252C%2522id%2522%253A%2522clsz0h67t00022e696gdauoqe%2522%252C%2522activeTabId%2522%253A%2522clsz0ieva00022e692uqb170o%2522%257D%252C%2522clsz0h67t00052e69c281fr1z%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clsz0h67t00042e69s4in6ui7%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522UNASSIGNED_PORT%2522%252C%2522port%2522%253A0%252C%2522path%2522%253A%2522%252F%2522%257D%255D%252C%2522id%2522%253A%2522clsz0h67t00052e69c281fr1z%2522%252C%2522activeTabId%2522%253A%2522clsz0h67t00042e69s4in6ui7%2522%257D%252C%2522clsz0h67t00032e69d0dlryqu%2522%253A%257B%2522tabs%2522%253A%255B%255D%252C%2522id%2522%253A%2522clsz0h67t00032e69d0dlryqu%2522%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Atrue%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D).

If you open the console, you'll see that the order of the print statements changes depending on whether you click Update or Toggle mount!

smoores-dev:
Done!

smoores-dev:
Oh, wait, maybe I should have checked; were you saying we should go ahead with approach #1 (where we never queue) or #2 (where we always queue unless we're unmounting)?

smoores-dev:
Ok now done!

tilgovi (approved):

tilgovi:
👏🏻

PR #107, opened 2024-02-23
Reorganize contexts and exports to match updates on main

smoores-dev:
This basically just splits up the LayoutGroup into its context, component, and hooks, like we have it on main! We were already directly using the EditorContext.Provider on this branch so I just got rid of the component that wraps it; we can try to break down the ProseMirror component like we have it on main if we'd like to later.


tilgovi (approved):
I didn't look tooooo closely! I assume you're trying to minimize the diff with main, and we can reconcile before any final merge.

PR #110, opened 2024-02-23
Decline version bump for ip upgrade

tilgovi:


tilgovi:
I'm seeing checks fail and I think it's maybe just because I merged the dependabot PR and that didn't include a version bump?

smoores-dev:
I don't think that would be the case. Once it's merged into main, I don't think yarn cares whether there's a version bump for it

tilgovi:
I'll rebase that branch on here and we'll see!

tilgovi:
😆 Doesn't like having multiple versioning files in the branch. I will... try again?

tilgovi:
Ah. I must have based it on a branch that no longer existed or something. All sorted.

PR #129, opened 2024-08-05
useEditorEventCallback should return value of callback

BrianHung:


tilgovi:
@smoores-dev has a similar PR at #118. I think in both cases this changes the return value of the callback to be `R | void`. Does TypeScript properly infer that, or do we need to add an explicit return signature to the definition of `useEditorEventCallback`?

BrianHung:
Ah, let's close in favor for #118. I'm on `typescript@4.9.5`. The inferred type is

```ts
function useEditorEventCallback<T extends unknown[], R>(callback: (view: EditorView, ...args: T) => R): (...args: T) => R | undefined
```

```
Returns a stable function reference to be used as an event handler callback.

The callback will be called with the EditorView instance as its first argument.

This hook is dependent on both the EditorViewContext.Provider and the LayoutGroup provider. It can only be used in a component that is mounted as a child of both of these providers.
```

tilgovi:
Cool, thanks for checking and for open the PR! I'll go merge #118.

tilgovi:
Published in v0.6.2.

PR #130, opened 2024-09-04
Better conversion from html attributes to react props

smoores-dev:
At the moment, there's a bug in the `OutputSpec` component that results in the ProseMirror output spec (result of `toDOM`) overwriting html attributes that come from node decorations. In particular, if the output spec specifies, e.g., `class: "paragraph"`, and a node decoration specifies `class: "empty-node"`, the `className` prop is set to `"paragraph"`. This is incorrect for two reasons:

1. The decoration spec should "win" over the output spec
2. For classes specifically, decoration specs and output specs should be merged, rather than overwriting one another

In general, we had a somewhat flimsy conversion from HTML attributes (as specified in output specs and decoration specs) to React props. So this PR introduces a new modules for converting HTML attributes to React props, as well as merging two sets of React props.

The conversion is based on React DOM's component props API: https://react.dev/reference/react-dom/components/common. I went through and, for any prop that has a different type or different name than its HTML analogue, added a case in the converter.

The merging is fairly simple: Props from the second of two sets supersede props from the first. For `className`, the two values are concatenated. For `style`, the two objects are merged.

Then I specifically updated `OutputSpec` to utilize this new props module


src/props.ts
```diff
@@ -0,0 +1,219 @@
...
+ * will be merged.
+ */
+export function mergeReactProps(
+  a: HTMLProps<HTMLElement>,
+  b: HTMLProps<HTMLElement>
+) {
+  return {
+    ...a,
+    ...b,
+    className: cx(a.className, b.className),
```
tilgovi, line 39: Is `cx` overkill here? Could we just do `${a.className} ${b.className}`?

src/props.ts
```diff
@@ -0,0 +1,219 @@
...
+ * that can be passed to a React component's
+ * `style` prop.
+ */
+export function cssToStyles(css: string) {
+  const cssJson = `{"${css
+    .replace(/;? *$/, "")
+    .replace(/;+ */g, '","')
+    .replace(/: */g, '":"')}"}`;
+
+  const obj = JSON.parse(cssJson);
```
tilgovi, line 19: Should we use the CSSOM for this?

```
const stylesheet = new CSSStyleSheet();
stylesheet.insertRule(`* { ${css} }`);
const style = stylesheet.cssRules[0].style;
```

Then, you should have a `CSSStyleDeclaration`. Iterating over that yields keys and you can get the property values and priorities.

src/props.ts
smoores-dev: `className` could be `undefined`, so it would be like slightly more than that; feels just clunky enough to use `cx` for haha

tilgovi: Fine for me.

src/props.ts
smoores-dev: Oooo yes, that's awesome, I like that much better

smoores-dev: Oh hm. `cssRules` returns a `CSSRuleList`; each item is a `CSSRule`, but I don't see a way to get a `CSSStyleDeclaration` from that? Weirdly, MDN says that "`document.styleSheets[0].cssRules[0].style` returns a `CSSStyleDeclaration` object", but that doesn't seem to follow from the types and the rest of the documentation...

tilgovi: `cssRules[0].style` returns a `CSSStyleDeclaration` in my browser.

smoores-dev: Hrmmmm. It works when I actually do it in the browser. So... something weird with the types, then. Maybe it's just that I need to discriminate on the `type` property. I'll figure it out!

smoores-dev: Ok cool. I ended up just casting it because:

(a) we're the one creating the rule, we know it's a style rule
(b) Apparently `CSSRule.type` is deprecated and that leaves me sort of baffled as to how I'm meant to detect the type anyway?

Do you think there are any performance concerns about doing this like... many many times during rendering? Should we just have one stylesheet, insert the rule, and then delete it again?

tilgovi: Having a single stylesheet seems like a reasonable optimization. No idea how expensive any of this is.

tilgovi: Don't worry about it, though. It looks like the constructor can run 4 million times a second on my machine.

smoores-dev: Hahah I was just setting up that same test. Great!

.eslintrc.json
```diff
@@ -17,6 +17,7 @@
   },
   "rules": {
     "no-console": ["error", { "allow": ["error"] }],
+    "react/prop-types": "off", // Typescript does this better
```
tilgovi, line 20: The rule's documentation talks about understanding TypeScript. What's complaining?

tilgovi (approved):

.eslintrc.json
smoores-dev: Mmmmm actually it looks like nothing, now. In a previous iteration of this, it was upset about my accessing a property on an object named `props` in `OutputSpec` (if I remember correctly, which I may not be, it wasn't even the actual component props, just another object created in the render function named `props`). Anyway, I'll undo this, since it's not needed anymore. If it comes up again we can evaluate

smoores-dev:
@tilgovi would you merge and release this for me? :)

tilgovi:
Done. `0.7.0-next.7` on the `next` tag.

PR #135, opened 2024-09-17
Support strict mode and add some memoization

### LINKED ISSUE #128: backspace does not work on first keystroke for new single character text block
This is using the latest react-editor-view branch. Sounds sort of similar some of the issues in #42

When a new paragraph or heading is added with a single character, backspace does not work. To remove the character backspace must be hit twice. Note this does not seem to be isolated to backspace, as shift left also fails to highlight the character.

* I am also using the baseKeyMap (for backspace) but I've validated none of the commands are used so this goes to the normal keystroke handling for the content editable.
* The first backspace does create a transaction but there are no steps associated with it
* Prosemirror document state is identical for both backspaces (one that works and one that doesn't).
* If I load the same state fresh in a new prosemirror instance backspace works fine, this only occurs in the editor when the single character node is created
* If I add more characters (or paragraphs after it) every works as expected

https://github.com/nytimes/react-prosemirror/assets/20848495/6c321086-ca14-48d7-b91b-aab084fdcf3e

smoores-dev:
Ok so! In order to resolve #128, and more generally support React's Strict Mode and concurrent rendering, we need to make it possible to re-render our components in isolation. Right now, in order to build the `ViewDescriptor` tree, we have to render the entire tree. This means that concurrent behaviors like re-using state and interrupting cause issues (the issue in #128 is that if a text node gets re-rendered in isolation, its view descriptor no longer has a correct reference to its parent).

This PR takes an approach to resolve this. We're having the reactKeys plugin do double duty; while it builds and tracks the map from positions to keys, it also tracks parent child relationships between keys.

Each node now registers its view descriptor with a single shared context, using its key. It can find its child and parent view descriptors by _their_ keys, using the new maps from the react keys plugin.

A neat side benefit of this is that, since it's now safe to render only parts of the tree at a time, we can memoize our components. This can lead to a pretty significant reduction in React rendering work; if typing at the end of the last paragraph in a document, none of the previous block nodes will re-render as you type.


smoores-dev:
@tilgovi I'm still working on this and it's breaking some tests at the moment,, but I'd still love your thoughts on it if you have any

.eslintrc.json
```diff
@@ -16,6 +16,7 @@
     "node": false
   },
   "rules": {
+    "react/prop-types": "off",
```
smoores-dev, line 19: We don't have to check this in, I'm just turning it off for new because using `React.memo` inline breaks this rule for some reason.

src/components/TextNodeView.tsx
```diff
@@ -1,32 +1,38 @@
...
 
   updateEffect() {
-    const { decorations, siblingDescriptors, node } = this.props;
+    const { decorations, viewDescContext, node, state, pos } = this.props;
+    const reactKeysState = state && reactKeysPluginKey.getState(state);
+    const key = reactKeysState?.posToKey.get(pos);
+    const parentKey = key && reactKeysState?.keyToParent.get(key);
+    const parent =
+      parentKey !== undefined ? viewDescContext[parentKey] : undefined;
+
```
smoores-dev, line 31: This is the actual fix to #128: when Strict Mode re-renders just the TextNodeView, we are still able to use the key map to find the parent view descriptor, so we can maintain the bidirectionality of the view descriptor tree

src/plugins/reactKeys.ts
```diff
@@ -1,15 +1,17 @@
-import { Node } from "prosemirror-model";
 import { Plugin, PluginKey } from "prosemirror-state";
 
 export function createNodeKey() {
   const key = Math.floor(Math.random() * 0xffffff).toString(16);
   return key;
 }
 
+const rootKey = "ROOT_KEY";
```
smoores-dev, line 8: This is just a stand-in; probably should be a Symbol or something.

src/components/ChildNodeViews.tsx
```diff
@@ -79,7 +80,10 @@ type SharedMarksProps = {
   childViews: Child[];
 };
 
-function InlineView({ innerPos, childViews }: SharedMarksProps) {
+const InlineView = memo(function InlineView({
```
smoores-dev, line 83: Most of the changes here are just wrapping components with `memo`

smoores-dev:
Ah, I forgot about MarkViews and WidgetViews haha

smoores-dev:
Yeah I don't think there's an easy way to make this work for mark and widget views. We get to sort of cheat with the actual React element keys, because mark views always wrap their child node views (and so can have the same keys, since keys only need to be unique at a given level of the tree) and widgets get given keys manually at construction.

But for this usage, we would need to know these keys during the state update (and without access to the decorations, necessarily), and they need to be a distinct set (meaning that mark keys need to be distinct from their child node keys). I don't think that's feasible.

I'm still not 100% sure that it really makes sense for us to worry about strict mode/concurrent rendering, aside from the annoying (but minor) bug in strict mode, #128. It may be enough to just say "You can't have a Suspense boundary within a ProseMirror doc". The memoization would have been nice to work out, and there may be a way to take some of this work and put it into changes that are only concerned with memoization.

PR #140, opened 2024-10-05
In a README examples, create the initial editor state only once

alexandresalome:
The code examples use react state with this example:

```javascript
const [state, setState] = useState(EditorState.create({ schema }));
```

With those examples, on every render, `EditorState.create` will be called.

To avoid that, we can use an initializer function ([documentation](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state)). In our case, we replace with:

```javascript
const [state, setState] = useState(() => EditorState.create({ schema }));
```


tilgovi (approved):
Thank you!

PR #144, opened 2024-11-30
Fix extra parenthesis in README.md

OwLu0905:
Fix extra closing parentheses in the useEditorEffect code examples in README.md.

The same typo appears twice in the SelectionWidget code example:
`const coords = view.coordsAtPos(view.state.selection.anchor))`

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ export function SelectionWidget() {
     if (!ref.current) return
 
     const viewClientRect = view.dom.getBoundingClientRect()
-    const coords = view.coordsAtPos(view.state.selection.anchor))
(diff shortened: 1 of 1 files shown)


smoores-dev (approved):

OwLu0905:
@smoores-dev  Thanks for merging. Before pushing I actually had executed `yarn fix`.
I noticed that `prettier` did not correctly write its changes to `README.md` at that time.
After investigating, I suspect this behavior might be caused by a race condition between `prettier` and `markdown-toc` when they're trying to modify `README.md` concurrently. This would explain why the formatting changes weren't properly applied.

PR #145, opened 2024-11-30
Fix formatting for README.md code blocks

smoores-dev:
#144 actually fixed formatting of the README.md file, which had previously been silently ignoring code blocks because of the syntax error that Prettier couldn't handle!

Now that Prettier can read these, they're failing the formatting checks. This just applies `yarn fix:format`!

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,30 +153,30 @@ we can use `useEditorEffect`.
 
 ```tsx
 // SelectionWidget.tsx
-import { useRef } from "react"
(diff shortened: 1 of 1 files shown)


smoores-dev:
We should also probably make it possible to run PR checks on PRs from non-maintainers, even if they don't run by default (though maybe we even just want to run them by default, since we don't have any GitHub secrets in this repo and the current scope of PR checks is very small)

tilgovi (approved):

PR #148, opened 2024-12-17
Enable SSR

smoores-dev:
I'm not 100% sure about this one. It's... got some hacks. Here's the idea:

A really cool use case for React ProseMirror is rich text editors that can share components between the editor and their reader-facing interface. One very simple way to implement this is to just render your ProseMirror document _with React ProseMirror_ in your reader interface, with editable set to false.

I think this is awesome, and folks like dskrpt.de and moment.dev are already doing this (see https://moment.dev/blog for an example!). It allows dskrpt to easily track the ProseMirror positions for highlights created by readers, for example, so they can be mapped through changes.

One significant disadvantage to this approach, though, is that you can't currently render the `<ProseMirror>` component on the server, because it creates a DOM element in its render function. Even if we got rid of that behavior, it would SSR to an empty document, because the initial construction of the EditorView happens in a layout effect, which doesn't run in SSR.

So this PR moves the initial EditorView creation into the render function of the ProseMirror component. In order to do this, we have to construct the EditorView without a mount initially, which ProseMirror handles by creating an empty div. To work around the EditorView constructor accessing `window` and `document`, we introduce some super dumb stubs for window and document that we set before calling `super` and then clean up (if we're in SSR).

The result is pretty cool -- you can render your ProseMirror document to HTML completely on the server. The only thing we defer is non-React node views, since those use Portals which only work on the client.


smoores-dev:
Also this is not a very big PR, I just thought I needed to update yarn to add a prelease version bump, buuut I think I didn't so I can pull that out into a separate PR. But that's why the diff count is so huge

PR #158, opened 2025-01-23
feat!: automatically use node view components

tilgovi:
Remove the need to call `useNodeView` and render the portals manually. Make it happen automatically it `<ProseMirror>`.

Remove `useNodeView` from the public API.

### DIFF
--- a/src/hooks/useNodeViews.tsx
+++ b/src/hooks/useNodeViews.tsx
@@ -14,7 +14,7 @@ import type {
 } from "../nodeViews/createReactNodeViewConstructor.js";
 
 export function useNodeViews(
-  nodeViews: Record<string, ReactNodeViewConstructor>
+  nodeViews?: Record<string, ReactNodeViewConstructor>
 ) {
   const [portals, setPortals] = useState({} as NodeViewsContextValue);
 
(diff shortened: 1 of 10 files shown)


tilgovi:
This change would constitute a breaking change. I'm considering resetting `main` to what's on `v1` right now and retargeting this against `main`.

tilgovi:
There's also part of me that's worried that @smoores-dev and I have talked through doing this before and decided not to for some reason I can't remember. If so, please help me remember!

tilgovi:
This PR also fixes #157 because it makes it so that the second, forced render after mounting the editor view includes the initial node view portals.

smoores-dev:
I _think_ that this was mostly a desire to avoid coupling things while we were still solidifying the API. That said, does this change as is still work with non-React nodeViews? It seems that we would want to expose two node view props, or distinguish between types of node views by inspecting their props (i.e., looking for a `component` property)

src/hooks/useNodeViews.tsx
```diff
@@ -47,17 +47,19 @@ export function useNodeViews(
...
+        createReactNodeViewConstructor(constructor, registerPortal),
+      ]
+    );
+    return Object.fromEntries(wrappedNodeViewEntries);
   }, [nodeViews, registerPortal]);
 
   return {
-    nodeViews: reactNodeViews,
+    nodeViews: wrappedNodeViews,
     renderNodeViews: () => <NodeViews portals={portals} />,
```
smoores-dev, line 63: This probably doesn't need to be a function anymore if it's internal! I honestly forget why it was, but I don't think it's providing any functionality

tilgovi:
> I _think_ that this was mostly a desire to avoid coupling things while we were still solidifying the API. That said, does this change as is still work with non-React nodeViews? It seems that we would want to expose two node view props, or distinguish between types of node views by inspecting their props (i.e., looking for a `component` property)

This change does exactly what you describe regarding distinguishing between node view types. It wraps all the node view constructors, but ends up wrapping it with the identity function if the node view constructor returns a node view without a component property.

smoores-dev:
Oh I see! I had missed the new branch in createReactNodeViewFunction. Awesome :)

smoores-dev (approved):

src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -136,27 +130,42 @@ export function findNodeKeyUp(editorView: EditorView, pos: number): NodeKey {
...
       node,
       editorView,
       getPos,
       decorations,
       innerDecorations
     );
 
-    let componentRef: NodeViewWrapperRef | null = null;
+    const { component: NodeView } = nodeView;
+    if (!NodeView) {
```
saranrapjs, line 152: worth adding a comment here to note that this is the branch which continues to allow specifying non-React nodeviews?

saranrapjs (approved):
This looks good to me! I am assuming (but have not run myself to confirm) that this passes the failing widget decoration test — is that worth adding here, too?

tilgovi:
> This looks good to me! I am assuming (but have not run myself to confirm) that this passes the failing widget decoration test — is that worth adding here, too?

I checked that it passed, but I'll commit it separately.

PR #162, opened 2026-04-01
Guard findNodeKeyUp against out-of-range positions

coolov:
## Issue
`findNodeKeyUp` calls `doc.resolve(pos)` without validating that pos is within the document bounds, causing a "RangeError: Position -1 out of range" crash when a node view's `getPos()` returns -1 (the standard sentinel for nodes that are no longer part of the document)

## Fix
This PR adds a bounds check `pos < 0` that returns `ROOT_NODE_KEY` early, if the node has no valid position, there are no ancestors to find

### DIFF
--- a/src/nodeViews/createReactNodeViewConstructor.tsx
+++ b/src/nodeViews/createReactNodeViewConstructor.tsx
@@ -94,7 +94,9 @@ export function findNodeKeyUp(editorView: EditorView, pos: number): NodeKey {
   const pluginState = reactPluginKey.getState(editorView.state);
   if (!pluginState) return ROOT_NODE_KEY;
 
-  if (pos < 0) return ROOT_NODE_KEY;
+  // -1 is the standard sentinel for nodes that are no longer part of the document
+  // since the node has no valid position, there are no ancestors to find
+  if (pos === -1) return ROOT_NODE_KEY;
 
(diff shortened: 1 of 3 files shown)


src/nodeViews/createReactNodeViewConstructor.tsx
```diff
@@ -94,6 +94,8 @@ export function findNodeKeyUp(editorView: EditorView, pos: number): NodeKey {
   const pluginState = reactPluginKey.getState(editorView.state);
   if (!pluginState) return ROOT_NODE_KEY;
 
+  if (pos < 0) return ROOT_NODE_KEY;
```
saranrapjs: can you add a comment here about the `-1` sentinel/why we have this check?

saranrapjs:
I think we use the yarn berry release process (e.g. `yarn version check -i`) to manage releases, but maybe this doesn't matter if we're aiming to publish this in the same step as we're making the change?

saranrapjs (approved):

src/nodeViews/createReactNodeViewConstructor.tsx
annabialas: agree, also we could make this slightly more self-explanatory by doing an equality check to -1. but that's a nit.

